PR #3, opened 2015-04-09
Add migration notice and optional jquery support

FagnerMartinsBrack:
@carhartl, should we:

* Force the user to rename the API upon migration, or
* Add the jQuery dependency, but make it optional and then remove it in [2.0 branch](https://github.com/js-cookie/js-cookie/tree/2.0)?


carhartl:
I think we should go for the second option.

FagnerMartinsBrack:
Ok

FagnerMartinsBrack:
Lucky we [didn't choose to return anything](https://github.com/carhartl/jquery-cookie/pull/330) for module loaders.

FagnerMartinsBrack:
@carhartl 

I've readded jQuery dependency and made it optional. Also I created some tests for AMD and nodejs module loaders.

Tell me what you think.

PR #5, opened 2015-04-10
Version 2.0

FagnerMartinsBrack:
This PR is a work in progress to keep track and discuss the version 2.0 features and removals.

- [x] Remove jquery dependency (https://github.com/js-cookie/js-cookie/commit/b74edfce74e074baab400129ab89084e59bfced9)
- [x] Remove the `raw` option (branch: `raw-option-removal`, #7)
- [x] Encode/Decode according to RFC 6265 (https://github.com/carhartl/jquery-cookie/issues/295) (branch: `RFC-6265`, #8)
- [x] Remove the `json` option (branch: `remove-json`, #9)
- [x] Allow Objects and Arrays to be stored as JSON (https://github.com/carhartl/jquery-cookie/pull/285) (branch: `auto-json-handling`, #10)
- [x] Remove all package managers code except from npm and bower (branch: `remove-package-managers`, #11)
- [x] Review and apply minification techniques to remove bytes (without sacrificing too much legibility) (branch: `few-tweaks`, #14)
- [x] Rework enclosed quotes rule (branch: `quoted-string`, #15)
- [x] Document default options to make clear how they work (branch: `default-behavior-docs`, #17)
- [x] Test how modern browsers handle characters outside ASCII range when sending cookies to the server (branch: `server-encoding-test`, #20)
- [x] Review the return value of the `remove` method (is it useful or worth it?) (branch: `void-remove`, #21)
- [x] Review the converters behavior to address the encoding/decoding issue (branch: `rethink-converters`, #22)
- [x] Highlight features in the README (branch: `features`, #23)


FagnerMartinsBrack:
@carhartl 

What about the return value of Cookies.remove. Is it worth returning something?

I'd rather turn into void and let the user test if the cookie was removed or not than doing something that is not useful for a majority of cases.

carhartl:
@FagnerMartinsBrack 

Yes, I agree, I don't see much use in it either, just seemed right™ to return such a value back then. 

Let's us simplify!

FagnerMartinsBrack:
I will be merging this in master while we determine the "path" behavior in #25. The master branch should be considered unstable anyway and the file is already marked as "2.0.0-pre" but holding reference to the 1.x docs.

PR #7, opened 2015-04-10
Remove encoding operation and raw option

FagnerMartinsBrack:
The encoding operation is going to be handled by the RFC 6265 implementation.

- [x] Remove code
- [x] Remove docs


carhartl:
Just thinking, so we maybe still need `raw`, with a slightly changed behavior, that is? If we handle JSON automatically, there may be occasions where someone would need to turn that off, and raw would be the global switch to turn that of, including. Also when some other encoding doesn't play nice with Rfc 6265...

FagnerMartinsBrack:
What about converters? Couldn't they handle everything raw option does currently?

carhartl:
Converters are supposed to be applied after the decoding. We cannot use them to undo decoding/json parsing.

carhartl:
To be honest, by now I believe converters have been a bad idea. Has anybody ever used them?

Nobody gains anything from this: ```
Cookies.get('foo', Number)
// vs
Number(Cookies.get('foo'))
```

That said, I'd rather remove converters as well for 2.0. What do you think?

FagnerMartinsBrack:
It makes sense, the spec only says:

> To convert the cookie-string
   (or components thereof) into a sequence of characters (e.g., for
   presentation to the user), the user agent might wish to try using the
   UTF-8 character encoding [RFC3629] to decode the octet sequence.
   This decoding might fail, however, because not every sequence of
   octets is valid UTF-8.

But it seems to be directed to the user agent and not how special cookie characters should be interchangeable after being processed by the user agent.

I remember there was that [PHP report](https://github.com/carhartl/jquery-cookie/pull/62) which encoded "+" sign for spaces instead of %x20 which would be its hex equivalent. I didn't see anywhere in the RFC stating which rule we should use to encode cookie values, it just say the values that are NOT allowed.

I am assuming we encode and decode using the UTF-8 hex equivalent :S.
Any idea how to handle this?

> To be honest, by now I believe converters have been a bad idea. Has anybody ever used them?

The only use I can think of converters is if somebody want to provide a different data type on cookies, but It seems JSON is generic enough...

Do we need an "override" converter to take place instead of the default one to prevent choosing which encoding to consider?

FagnerMartinsBrack:
Going back to the `raw` option, I see the problem is all about the malformed cookies with characters that are not allowed in the value right? what if we provide a way to override the default encoding/decoding mechanism? Then we could implement a custom encoding of special characters and ppl could override it to change the default. That would solve the problem and make the "raw" option obsolete. The only problem seems to be the inconsistent encoding of characters that are not allowed in the cookie value, like [PHP](https://github.com/carhartl/jquery-cookie/pull/62).

FagnerMartinsBrack:
Look at the last commits, I provided 2 examples of server-side cookies with a custom decoding function using converters. I believe this accounts for all issues reported so far regarding third party cookies...

PR #8, opened 2015-04-11
RFC 6265 compliance

FagnerMartinsBrack:
- [x] cookie-value
- [x] cookie-name
- [x] docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,13 @@ Cookies.remove('name', { path: '/' }); // => true
 
 *Note: when deleting a cookie, you must pass the exact same path, domain and secure options that were used to set the cookie, unless you're relying on the default options that is.*
 
+## RFC 6265
+
+This project assumes [RFC 6265](http://tools.ietf.org/html/rfc6265#section-4.1.1) as a reference for everything. That said, some custom rules are applied in order to provide robustness and cross-browser compatibility.
+
+### Encoding
+All special characters that are not allowed in the cookie-value or cookie-name in at least one supported browser are encoded/decoded with each UTF-8 Hex equivalent. Special characters that consistently work among all supported browsers are not encoded/decoded this way.
+
 ## Configuration
 
 ### json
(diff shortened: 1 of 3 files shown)


carhartl:
:tada:

FagnerMartinsBrack:
:D

I still need to test if using `encodeURIComponent/decodeURIComponent` is necessary for UTF-8 chars or if the browser and the servers can handle it accordingly. If not then need to discuss whether this should be handled by js-cookie or by the user.

It says (emphasis mine):

```
NOTE: Despite its name, the cookie-string is actually a sequence of
   octets, not a sequence of characters.  To convert the cookie-string
   (or components thereof) into a sequence of characters (e.g., for
   presentation to the user), the >>>>USER AGENT<<<< might wish to try using the
   UTF-8 character encoding [RFC3629] to decode the octet sequence.
   This decoding might fail, however, because not every sequence of
   octets is valid UTF-8.
```

carhartl:
I think said conversion itself is handled by the client internals and it will never make it into `document.cookie`. So we only need to make sure the characters that aren't allowed are decoded properly?

FagnerMartinsBrack:
That's my theory, but still I believe it's worth setting up a custom server to test exactly each browser's behavior in this matter. The spec unfortunately says:

> ... might wish to try using the UTF-8 character encoding ...

"might" might not be enough T.T

We still need some time for ppl to be aware of js-cookie, I believe there is no rush for a 2.0 release unless somebody requests so on npm. I am thinking about throwing some betas after js-cookie awareness is higher to see if there's any issues in the wild.

carhartl:
Agreed. Btw, I was thinking, why are we naming it 2.0? Since it will be rather detached from jquery-cookie, we could just as well version it as 1.0?

FagnerMartinsBrack:
Wouldn't that confuse the tags system? https://github.com/js-cookie/js-cookie/releases

carhartl:
Yes, right, at least it will confuse everyone (one can push any tag any time).

PR #10, opened 2015-04-13
Add JSON automatic handling

FagnerMartinsBrack:


README.md
```diff
@@ -72,6 +72,28 @@ Cookies.remove('name', { path: '/' }); // => true
 
 *Note: when deleting a cookie, you must pass the exact same path, domain and secure options that were used to set the cookie, unless you're relying on the default options that is.*
 
+## JSON
+
+js-cookie provides automatic JSON storage for cookies.
+
+When creating a cookie you can pass an Array or Object Literal instead of a string in the value. If you do so, js-cookie store the string representation of the object according to the `JSON.stringify` api (if available):
```
carhartl: Strictly technical, it doesn't need to be an Array/Object literal, does it? E.g.
```js
Cookies.set("foo", new Array("foo"))

In that case I'd rather write something like "...you can pass an instance of Array or Object..."

**Comment by carhartl** [comment]
@FagnerMartinsBrack 

I am really not happy about `getJSON()`, was it necessary? I thought the plan was to handle that automatically, within the same reader method...?

Unfortunately you didn't give me enough time to review beforehand.

**Comment by carhartl** [comment]
You're too fast for me :)

File: README.md (hunk 1 above)
Comment (FagnerMartinsBrack) [inline README.md]: Then we would be required to support it right? I don't see a clear reason for using an Array instance, if we were to support that then in that case we should for consistency also support Object instance, but Object can be anything, it can be a Date instance, a String instance, a custom object with methods without any relationship with key:value pairs, etc.

We could document that an "Array instance" is allowed, in this case it would work for Array constructors and literals, but then would we also support Objects for consistency? Do you think this is bad because ppl might read "Literals" as objects that should not be stored in variables? The intent here is just to limit the creation of the instances using literal notations, which AFAIK already handles all use-cases.

Comment (FagnerMartinsBrack) [inline README.md]: We could document that if it's not a valid JSON structure then it will use `String(object)` in the input, but then we would be enforcing behavior for invalid input, and any changes that does something else would be backwards incompatible.

Assuming we are using [semver spec](semver.org), backwards incompatible changes should only be  release in major version (in this case v3.0).

Comment (carhartl) [inline README.md]: > but Object can be anything, it can be a Date instance, a String instance, a custom object with methods without any relationship with key:value pairs, etc.

Now I get your point. I was specifically remembering that once I needed to fix an issue so this works:

```js
$.cookie("foo", new String("..."))
```

Than I thought, if we say "literals" we might be unnecessarily restricting users to use literals. I didn't think of all the other objects. I just thought `new Array` and `new Object` will probably work just as well...

FagnerMartinsBrack: So `When creating a cookie you can pass an array or plain object instead of a string`?

carhartl: Yes, that's use that!

FagnerMartinsBrack:
X-Ref: https://github.com/carhartl/jquery-cookie/pull/285#issuecomment-93530332
X-Ref: https://github.com/js-cookie/js-cookie/pull/5#issue-67595065

PR #12, opened 2015-04-15
Make it clear js-cookie is supposed to work in any relevant browser

FagnerMartinsBrack:
The build status matrix just shows which browsers we test against

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ A simple, lightweight JavaScript API for handling cookies
 
 ## Build Status Matrix
 
+js-cookie is intended to work in any relevant browser. For the sake practicality though, we test only against browsers that have a considerable market share.
+
 [![Selenium Test Status](https://saucelabs.com/browser-matrix/js-cookie.svg)](https://saucelabs.com/u/js-cookie)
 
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
whatever

carhartl:
Maybe add a simple list of browsers we intend to to the documentation? Or we assume that people deduct support from the build matrix...

FagnerMartinsBrack:
I started this as an idea to document each browser support, but then realized the build status matrix already does that (we can't support what we don't test right?).

When I was writing the notice in this PR I realized the "we test only against browsers that have a considerable market share" could have a negative impact. What is a considerable market share? People might look at that and immediately leave to try something else assuming we don't support IE8, for instance (even though it's discretly shown in the matrix).

When I realized I was thinking too much and not going anywhere I gave up and went to do something else xD

We can think about it though, which browsers support, which browsers test, etc.

carhartl:
Hmm, yeah, let's point to the build matrix :)

I've added an issue to update/enhance it.

FagnerMartinsBrack:
:+1:

PR #14, opened 2015-04-16
Few tweaks and tricks to reduce gzipped size while supporting IE8

FagnerMartinsBrack:
https://github.com/js-cookie/js-cookie/commit/78a3b44b82c997b9b682af1b34b72a664f8cc563:
```shell
raw     gz Sizes
1567    838 build/js.cookie-2.0.0-pre.min.js
3859   1494 src/js.cookie.js

raw     gz Compared to 2.0 @ 2847ee7906960acfe25821284b24ead5d697c4c3
-91    -54 build/js.cookie-2.0.0-pre.min.js
-182   -89 src/js.cookie.js

raw     gz Compared to master
+14    +13 build/js.cookie-2.0.0-pre.min.js
+132   -57 src/js.cookie.js
```

https://github.com/js-cookie/js-cookie/commit/d02fac4fb932eb538860bfad2997199ec16c212d (after the merge of `quoted-string` branch):

```shell
raw     gz Sizes
1525    820 build/js.cookie-2.0.0-pre.min.js
3747   1432 src/js.cookie.js

raw     gz Compared to 2.0 @ 2847ee7906960acfe25821284b24ead5d697c4c3
-133    -72 build/js.cookie-2.0.0-pre.min.js
-294    -151 src/js.cookie.js

raw     gz Compared to master
-28    -5 build/js.cookie-2.0.0-pre.min.js
+20    -119 src/js.cookie.js
```

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -46,19 +46,8 @@
 		return value;
 	}
 
-	function processWrite (value) {
-		var stringified;
-		try {
-			stringified = JSON.stringify(value);
-			if (/^(?:\{[\w\W]*\}|\[[\w\W]*\])$/.test(stringified)) {
-				value = stringified;
-			}
-		} catch(e) {}
-		return encode(String(value), unallowedCharsInValue);
-	}
-
 	function processRead (value, converter, json) {
-		if (value.indexOf('"') === 0) {
+		if (value.charAt(0) === '"') {
 			// This is a quoted cookie as according to RFC2068, unescape...
 			value = value.slice(1, -1).replace(/\\"/g, '"').replace(/\\\\/g, '\\');
 		}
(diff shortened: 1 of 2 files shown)


src/js.cookie.js
```diff
@@ -71,64 +60,72 @@
...
+		}
+
+		// To prevent the for loop in the first place assign an empty array
+		// in case there are no cookies at all. Also prevents odd result when
+		// calling "get()"
+		var cookies = document.cookie ? document.cookie.split('; ') : [];
+		var i = 0;
 
-		for (; i < l; i++) {
+		for (; i < cookies.length; i++) {
```
FagnerMartinsBrack: @carhartl 

Unless somebody present a benchmark in which there's a difference of caching the length for less than 10k cookies in modern browsers, caching the length doesn't make sense. This has around ~4b gzip reduction and I doubt someone will have more than 10k cookies on the system in a way this causes some bottleneck. Modern engines are supposed to optimize this.

carhartl: Fully agree. If we decide not to support some ancient Internet Explorer, we might be able to return to using a regex. I had been resorting to a loop only because of a particular bug in IE...

FagnerMartinsBrack: Well, we are testing in IE8+, but I am still not trying to use features that break IE7-, like [this line](https://github.com/js-cookie/js-cookie/pull/14/files#diff-b686f7b30e7b0c4f5ed6cd2e87241704R50) which uses `charAt()` instead of array notation.

If you think it's ok to drop IE 7- then we can rewrite that and save even more bytes.

carhartl: No problem dropping IE 7. It think the problem existed in IE 8 though, need to look that up in the jquery-cookie repo.

FagnerMartinsBrack: Hmm carhartl/jquery-cookie#25, need to take a look on that.

carhartl: It's IE 8 as well: https://github.com/carhartl/jquery-cookie/issues/39

I just noted anyway, back then we didn't have the feature to return all the cookies when passing no arguments... I don't think er Van tackle that using the (same) regular expression. 

Is IE 8 still a thing?

carhartl: And who knows, maybe the implementation didn't even change in IE 9.

FagnerMartinsBrack: I will check that in a new PR: https://github.com/js-cookie/js-cookie/issues/19

FagnerMartinsBrack: https://github.com/js-cookie/js-cookie/issues/19#issuecomment-95418381

PR #15, opened 2015-04-16
Enclosed quotes

FagnerMartinsBrack:
@carhartl 

There's no mention of a slash to escape quotes in RFC 6265, is it safe to assume it's not going to be served from the server?
I am just expecting compliance from the server with `*cookie-octet / ( DQUOTE *cookie-octet DQUOTE )`. Opera doesn't allow plain quotes inside quoted cookie-value:

```shell
$ document.cookie = 'c="v"'
> 'c="v"'
$ document.cookie = 'c="v"2"'
> ''
```

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -48,8 +48,7 @@
 
 	function processRead (value, converter, json) {
 		if (value.charAt(0) === '"') {
-			// This is a quoted cookie as according to RFC2068, unescape...
-			value = value.slice(1, -1).replace(/\\"/g, '"').replace(/\\\\/g, '\\');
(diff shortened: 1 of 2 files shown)


carhartl:
I think so. Isn't a double quote simply not allowed?

  > US-ASCII characters excluding CTLs,
       ; whitespace DQUOTE, comma, semicolon

FagnerMartinsBrack:
It's not allowed inside a cookie-octet, but it's allowed inside a cookie-value (as long as it is the first and the last character):

```shell
cookie-value = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
```

PR #20, opened 2015-04-19
Test how browsers handle encoding when communicating with a server

FagnerMartinsBrack:


FagnerMartinsBrack:
IE 11 doesn't transmit 3-byte characters and above:

image

Safari 5.1.7 and PhantomJS doesn't transmit any 2-byte characters and above (only ASCII):

image

It seems the ["might"](https://github.com/js-cookie/js-cookie/pull/8#issuecomment-94158741) is not there just because.

FagnerMartinsBrack:
I was wrong after all. The browser doesn't encode or decode anything out-of-the-box when it communicates with the server, what it does is prevent a cookie to be created in the client if it doesn't comply with the characters the browser accepts.

Safari and IE still doesn't accept crazy chars in name/values, so we still need to stick with encoding everything. The difference now is that there's some valid characters in the spec encoded by `encodeURIComponent` that can be transmitted without being encoded at all. Need to treat those exceptions manually.

carhartl:
Interesting find!

FagnerMartinsBrack:
The problem is that we crash again into the problem of ppl having the `%` character in the value which makes decodeURIComponent throw a malformed error.

Here is how I "solved" this without using a `raw` option:

* When js-cookie decodes a cookie, it strips all valid hex encoded characters without messing with the rest of the value.
* Then it passes each stripped encoded characters through `decodeURIComponent` and replaces the result in the original value.

This prevents that a `%` character in the value breaks the decoding routine.

@carhartl [review this](https://github.com/js-cookie/js-cookie/blob/694ffe1e62b1eae13f6e7e3a00ccae950698e8a6/src/js.cookie.js#L34-L39) and tell me what you think.

**What if there is a value that matches a valid hex encoded string but should be treated as plain? Let's say a %20?**

Well, if it was written by js-coookie it is not possible, because it will encode the percent character using `encodeURIComponent` by default.

**What if it was written in the server using an arbitrary encoding mechanism?**

Then Houston, we have a problem.

* Should we provide a way to treat the cookie value before it's processed by `decodeURIComponent`?
* Should we make it possible to override `decodeURIComponent` with a custom function for decode?
* Should we make it possible to use a custom function for decoding AND encoding, making the `*URIComponent` the default implementation but making it possible to provide a custom one?

I agree that at this point converters are indeed useless as is. What we need here might be something to process the value before encode and/or decode.

carhartl:
Wait, the whole idea of going for the RFC conforming encoding** was to not use decodeUriComponent at all any longer (thus no longer running into said error), i.e. not having to use it. 

** maybe only a simplified version, in that we would encode at least those characters with a special meaning (";" etc.)

FagnerMartinsBrack:
> maybe only a simplified version, in that we would encode at least those characters with a special meaning (";" etc.)

But that doesn't work as I demonstrated in the first comment. Safari and IE does not set cookies with characters like `ã`, `₯` or `𩸽`. It's not practical to encode each character manually.

carhartl:
Yeah, I am not able to grasp the entire context/problem when only being able to take quick glances. 

Replacing decodeURIComponent was at least what I had in mind when adding https://github.com/carhartl/jquery-cookie/issues/295

From that point of view I don't yet see what we gain when conforming to the RFC on top of that.

FagnerMartinsBrack:
Well, the gain here is that unnecessary characters are not encoded, for example:

sharp (#) is allowed in a cookie value, but `encodeURIComponent('#')` equals to `'%23'`. So we need to prevent `encodeURIComponent` from encoding that character.

The RFC accepts the following characters:

```
%x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
```

which means

```
0x21: !
0x23-2B: #$%&'()*+
0x2D-3A: -./0123456789: 
0x3C-5B: <=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[
0x5D-7E: ]^_`abcdefghijklmnopqrstuvwxyz{|}~
```

It doesn't say anything for `ã`. Safari fails to set that character.

Let's say how `encodeURIComponent` handles those:

```
$ encodeURIComponent('!#$%&\'()*+-./0123456789: <=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~')
> "!%23%24%25%26'()*%2B-.%2F0123456789%3A%20%3C%3D%3E%3F%40ABCDEFGHIJKLMNOPQRSTUVWXYZ%5B%5D%5E_%60abcdefghijklmnopqrstuvwxyz%7B%7C%7D~"
```

which means the following characters are encoded by encodeURIComponent but shouldn't:

```
$ encodeURIComponent('#$%&+/0123456789: <=>?@[]^`{|}')
> "%23%24%25%26%2B%2F0123456789%3A%20%3C%3D%3E%3F%40%5B%5D%5E%60%7B%7C%7D"
```

I have created [some tests for this](https://github.com/js-cookie/js-cookie/blob/5235b01b6cb83a372df8430087889ef062202c1c/test/encoding.js#L58-L196), and [the implementation](https://github.com/js-cookie/js-cookie/blob/5235b01b6cb83a372df8430087889ef062202c1c/src/js.cookie.js#L85-L86).

Bottom of line: Encoding is necessary., but we can control which characters are encoded and which ones aren't.

FagnerMartinsBrack:
In the `cookie-name` the [allowed characters](http://tools.ietf.org/html/rfc2616#section-2.2) are "any US-ASCII character (octets 0 - 127)" except those specified as "CLT" or "separators". CLT are control characters that nobody will have in cookies, so we need to treat the "separators".

Let's see how `encodeURIComponent` deals with the "separators" characters that are not allowed:

```
$ encodeURIComponent('()<>@,;:\\"/[]?={} \t');
> "()%3C%3E%40%2C%3B%3A%5C%22%2F%5B%5D%3F%3D%7B%7D%20%09"
```

Hmm, everything is encoded except those parens there in the left, so we use `escape` to encode just the parens:

```javascript
```

Now what about the characters that are allowed in the `cookie-name` but encoded by `encodeURIComponent`? I came up with these:

```
$ encodeURIComponent('#$&\+^`|');
> "%23%24%26%2B%5E%60%7C"
```

So we decode them after using `encodeURIComponent` in the key:

```javascript
key = key.replace(/%(23|24|26|2B|5E|60|7C)/, decodeURIComponent);
```

Ok, but why not doing something like:

````javascript
String(key).replace(/[^#$&\+^`|]/g, encodeURIComponent)
```

?

Well...

![image](https://cloud.githubusercontent.com/assets/835857/7232875/56d6c432-e754-11e4-8574-2ffd5b5f47f5.png)

It fails for 4 bytes characters T.T

FagnerMartinsBrack:
Merging this in order to start new pull requests on top of branch 2.0 without code conflicts. Discussion can continue.

PR #22, opened 2015-04-20
Rethink converters

FagnerMartinsBrack:
Inspired by https://github.com/carhartl/jquery-cookie/issues/151#issuecomment-25108455

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ Cookies.getJSON(); // => { name: { foo: 'bar' } }
 
 This project is [RFC 6265](http://tools.ietf.org/html/rfc6265#section-4.1.1) compliant.  
 All special characters that are not allowed in the cookie-value or cookie-name are encoded with each UTF-8 Hex equivalent.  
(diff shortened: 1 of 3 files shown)


carhartl:
Great! Just wondering, what about the following alternative syntax:

```js
Cookies(function () { /* the converter function */ })
```

Maybe it's nonsense?

FagnerMartinsBrack:
In this case how the reader of the code would know the function represents a converter? The user could use:

FagnerMartinsBrack:
In this case how the reader of the code would know the function represents a converter?

Ok, it could be written as:

```javascript
Cookies(function converter() {
  /* ... */ 
});
```

But then in the `unescape` case things get cryptic.

```javascript
// What does unescape do here?
var cookies = Cookies(unescape);
```

FagnerMartinsBrack:
Merging this in order to start new pull requests on top of branch 2.0 without code conflicts. Discussion can continue.

PR #24, opened 2015-04-21
Add noConflict method to API

tomsouthall:
One issue with moving away from a jQuery plugin is that jQuery handles potential namespace conflicts very nicely using its `.noConflict` function.

Using `Cookies` as the namespace is fine if the script is being run on your own site. However if it needs to be run on a third party site, there is always the potential for namespace clashes, for example with [other cookie libraries](https://github.com/ScottHamper/Cookies) that also use `Cookies` as a namespace.

This adds a lightweight `.noConflict` method to the API, allowing the namespace to be reassigned and preserving the original value of `Cookies`.

The test suite and README are updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,38 +34,38 @@ The plugin can also be loaded as AMD or CommonJS module.
 
 ## Usage
 
-Create session cookie:
+**Create session cookie:**
 
 ```javascript
 Cookies.set('name', 'value');
 ```
 
-Create expiring cookie, 7 days from then:
+**Create expiring cookie, 7 days from then:**
 
 ```javascript
 Cookies.set('name', 'value', { expires: 7 });
 ```
 
-Create expiring cookie, valid across entire site:
+**Create expiring cookie, valid across entire site:**
 
 ```javascript
 Cookies.set('name', 'value', { expires: 7, path: '/' });
 ```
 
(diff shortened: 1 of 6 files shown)


FagnerMartinsBrack:
Definitely :+1:

I would try not to make too many change in master because in the branch `2.0` there's a lot of things being rewritten and that may cause heavy conflict. Do you see this as immediately necessary or it can be punted until version 2.0 is out?

FagnerMartinsBrack:
I suppose `.noConflict()` should not exist in `AMD` or `CommonJS` environments, there's no possibilities of conflict there because the script is loaded on-demand (and not registered globally).

For that reason I made some small modifications to expose the `noConflict` method only when the `Cookies` namespace is set globally (not just when there's no `window` present since you can declare a `window` variable in the parent scope prior to loading the script).

Take a look and tell me if that solves the issue: https://github.com/js-cookie/js-cookie/commit/a105e27b63eb831a8ded138f7d77b784e64cf7c1

tomsouthall:
@FagnerMartinsBrack thanks for the quick response and yes you are right about AMD CommonJS having no need of `.noConflict`. Your re-work makes complete sense. Great stuff.

PR #34, opened 2015-05-04
Remove use of deprecated ‘escape’ function

andraaspar:
According to MDN, [escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) is deprecated and should not be used.

I have replaced it with a lambda function. To somewhat reduce the increase in code size, I have also removed a few unnecessary backslashes from RegEx.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ node_modules
 build
 .sizecache.json
 *.log*
+# Eclipse project files
+/.settings/
+/.project
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -60,7 +60,7 @@
 
 				key = encodeURIComponent(String(key));
 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
-				key = key.replace(/[\(\)]/g, escape);
+				key = key.replace(/[()]/g, function (s) { return '%' + s.charCodeAt(0).toString(16); });
 
 				return (document.cookie = [
 					key, '=', value,
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -50,7 +50,7 @@
 
 				try {
 					result = JSON.stringify(value);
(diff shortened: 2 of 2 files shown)


src/js.cookie.js
```diff
@@ -60,7 +60,7 @@
 
 				key = encodeURIComponent(String(key));
 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
-				key = key.replace(/[\(\)]/g, escape);
+				key = key.replace(/[()]/g, function (s) { return '%' + s.charCodeAt(0).toString(16); });
```
FagnerMartinsBrack: Thanks for the PR.

Yes, `escape` is deprecated an is discouraged to be used to escape URIs. In this case we are not using to escape URIs, we are just using as a convenience to encode the parens and save bytes, the same goes for the uses of `decodeURIComponent` and `encodeURIComponent` throughout the code.

The web is pretty stable, functions like `escape` are historically implemented consistently and are not going to be removed by browser vendors. The amount of bytes it saves is worth the use even though it's deprecated, we have a serious challenge here that is keeping the code base in the ~800 bytes range. If there's a problem in the functionality of the script that is solved by not using escape, then we will gladly change it.

.gitignore
```diff
@@ -2,3 +2,6 @@ node_modules
 build
 .sizecache.json
 *.log*
+# Eclipse project files
+/.settings/
+/.project
```
FagnerMartinsBrack: Hey I also use eclipse and love it ^^. In this case there's no need to store the IDE specific files in the source repository, it is discouraged since it adds unnecessary verbosity for those that do not use eclipse. You can ignore the eclipse's dot files [globally](https://help.github.com/articles/ignoring-files#create-a-global-gitignore), in this case you can work in the project without worring about unmodified files being marked to be staged by EGit.

andraaspar:
OK, no worries.

PR #41, opened 2015-05-24
Add Composer support

RobLoach:
This adds support for [Composer](http://getcomposer.org).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,13 @@
+{
+  "name": "js-cookie/js-cookie",
+  "description": "A simple, lightweight JavaScript API for handling cookies",
+  "type": "component",
+  "license": "MIT",
+  "extra": {
+    "component": {
+      "scripts": [
+        "src/js.cookie.js"
+      ]
+    }
+  }
+}


FagnerMartinsBrack:
> Regarding your worry of un-maintained package management, Packagist does the release management for us, which makes your job a non-issue.

The worry I meant is having to maintain the "scripts" attribute if the location of the main file is changed.

Is it strictly necessary to add support for composer to make it work? Isn't there another way?

@kalenjordan reported in 2013 that he found a way to make it work without having to add another file:

> Actually realized that I can manage jquery-cookie under composer without needing to publish to packagist, so I can likely manage this myself without needing any update to this repository.

\- https://github.com/carhartl/jquery-cookie/issues/220#issuecomment-25412945

Not sure if the solution he found applies to everyone.

I will leave this up for @carhartl to decide.

kalenjordan:
@FagnerMartinsBrack wow it's been a while.  I don't remember off the top what the solution was or even what project this was in reference to.  I'm pretty sure I just found jquery-cookie on packagist and pulled it in as if it were a php dependency.  Sorry that I'm so foggy on the details.

carhartl:
@FagnerMartinsBrack 

I am a bit reluctant in terms of adding (back) in a variety of package managers, just after having removed a few.

So my suggestion is to wait and see whether the demand is rising for the proposed package manager. I can't tell how exotic or not that package manager is.

@RobLoach 

Is there a way to make packaging with composer work without the composer.json?

FagnerMartinsBrack:
@carhartl 

> I am a bit reluctant in terms of adding (back) in a variety of package managers, just after having removed a few.
> So my suggestion is to wait and see whether the demand is rising for the proposed package manager. I can't tell how exotic or not that package manager is.

I agree. Need more info about it.

X-ref: https://github.com/js-cookie/js-cookie/pull/11

drAlberT:
:+1:  for composer

FagnerMartinsBrack:
@drAlberT thanks for the feedback.

For now I will be closing this and assume that people using composer can implement npm/bower or handle support somewhere else. We will keep js-cookie strictly "JS" focused and avoid supporting package managers of other languages.

As @carhartl pointed out, we will keep watching the demand.

FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/issues/172#issuecomment-209286184 for detailed explanations of how to make js-cookie working with composer.

FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/issues/172#issuecomment-219906692, somebody from the community added this to Packagist.

Issue #38, opened 2015-05-13
Make it available on cdnjs.com


stephenson:
Would be great if js-cookie was available on cdnjs.com, read more here: https://github.com/cdnjs/cdnjs#adding-a-new-or-updating-an-existing-library

FagnerMartinsBrack:
Yeah, I am trying to figure that out: https://github.com/cdnjs/cdnjs/issues/4627

stephenson:
Fantastic!

FagnerMartinsBrack:
@stephenson it seems that issue is not going anywhere. I believe it's ok to add a new project named `js-cookie` in cdnjs. Do you know how to easily add [all tags](https://github.com/js-cookie/js-cookie/releases) to cdnjs without having to send one pull request for each? Do you want to help on this one? :)

FagnerMartinsBrack:
I started a PR to add js-cookie as a new lib: https://github.com/cdnjs/cdnjs/pull/4713

FagnerMartinsBrack:
js-cookie is available on cdnjs: http://cdnjs.com/libraries/js-cookie

stephenson:
Fantastic, thanks!

FagnerMartinsBrack:
I have added CDNJS as a community supported CDN. See https://github.com/js-cookie/js-cookie/wiki/Community-Supported-CDNs/262a9ae0cd09d911b02c3abaf6f3100f00feb530.

FagnerMartinsBrack:
[Added jsDelivr CDN to the README](https://github.com/js-cookie/js-cookie/pull/371#issuecomment-340131636) because now they support on-demand minification.

PR #46, opened 2015-06-08
Document the CDN

FagnerMartinsBrack:
@carhartl 
What you think about documenting the CDN in the "installation" section?

Pros:
* People might rely less on `src/js.cookie.js` file on master that may be unstable
* Encourages the use of CDN for general sites instead of downloading the unminified source
* It provides a documented way to download the minified version instead of having to upload on each release (there were several requests in the past to provide the minified version)

Cons:
* It's hard to maintain the link consistently on each tagged release. cdnjs picks a new version **after** a tag is pushed[1]. If we rely in that, the tag view will retain the link of the previous version instead of the new one.

*[1]: I didn't tested if this is working, it is just how cdnjs documents the ["autoupdate" feature](https://github.com/cdnjs/cdnjs/tree/31eee715a9ffa7a6e739da45b7083c0b0dcbc266#enabling-gitrecommended-or-npm-auto-update) they rely upon and how I [set up](https://github.com/cdnjs/cdnjs/blob/31eee715a9ffa7a6e739da45b7083c0b0dcbc266/ajax/libs/js-cookie/package.json#L16-L23) js-cookie on cdnjs.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ A simple, lightweight JavaScript API for handling cookies
 Include the script (unless you are packaging scripts somehow else):
 
 ```html
-<script src="/path/to/js.cookie.js"></script>
+<script src="//cdnjs.cloudflare.com/ajax/libs/js-cookie/2.0.0-beta.1/js.cookie.min.js"></script>
 ```
 
 **Do not include the script directly from GitHub (http://raw.github.com/...).** The file is being served as text/plain and as such being blocked


FagnerMartinsBrack:
@carhartl?

carhartl:
I think this is a good idea, but that a tag would contain an outdated link is why I in the end would not want to merge it.

carhartl:
It should be possible to automate this somehow in order to pull in the current version, at build time...

FagnerMartinsBrack:
@carhartl 

> It should be possible to automate this somehow in order to pull in the current version, at build time...

Any idea how to do it?

As it currently stands, the major problem is that the resource will be unavailable (`404`) when the tag is created. It is necessary to wait cdnjs to pick up the new version after the new tag is pushed. The problem is the same even if we find a way to generate the link dynamically.

Maybe we can create the tagged release first with the outdated link (that is working), then after the new version is ready on cdnjs, create a new commit and change the tag pointer to the new version.

It's too complicated, there has to be an easier way to do this. I wonder if it's worth just to provide a link to https://cdnjs.com/libraries/js-cookie in the README and be done with it :tired_face: 

What do you think?

carhartl:
I was thinking (naively) we could use https://github.com/geddski/grunt-release to automate the entire release process and along with it tackle updating the cdnjs reference. Though I haven't fully understood how the update process works at all with cdnjs...

FagnerMartinsBrack:
> Though I haven't fully understood how the update process works at all with cdnjs...

Basically it works like this:

1. [Send a PR to cdnjs containing a file called `package.json` with the project meta data](https://github.com/cdnjs/cdnjs/pull/4713).
2. [The `package.json` should have the `autoupdate` property, so the data inside the `autoupdate` will be used for cdnjs to autoupdate the new version whenever a new tag is pushed in the target repository (so I hope)](https://github.com/cdnjs/cdnjs/pull/4713/files#diff-0f4cc8ef43a762b1e8a204882d3fe372R16).
3. [To add a library for the first time, create a directory with each previous file versions in the format `ajax/libs/js-cookie/x.x.x/js.cookie.min.js`, where `x.x.x` is one of all the previous versions of the library](https://github.com/cdnjs/cdnjs/tree/6ea9043f37ec548071e625ae729859935e99dec1/ajax/libs/js-cookie).
3. [After a few hours, the page is created according to the `package.json` information using the versions submitted in the first PR](http://cdnjs.com/libraries/js-cookie).

And that's it.

I may take a look in `grunt-release`, but I have failed to use it in previous projects. Tools that promises to abstract the release process do not work for all cases, except for very specific ones (and the ones who work for all cases have so many configs that you're better writing your own).

FagnerMartinsBrack:
I found `jquery.cookie` is also in [jsDelivr](https://github.com/jsdelivr/jsdelivr/tree/master/files/jquery.cookie) CDN.

Even though it is available in other CDNs, if we decide to land any CDN documentation in this repository, it has to be cdnjs, the reasons are:

1. jsDelivr doesn't seems to be more popular than cdnjs (1k stars against 2k stars)
2. We only got a request to add cdnjs (#38)

FagnerMartinsBrack:
Closing this in favor of #57

PR #47, opened 2015-06-14
Testing if the build status matrix shows only for the master branch

FagnerMartinsBrack:
This PR is just to test if the build status matrix is really for the "master branch" as specified in the README. My assumption is that "https://saucelabs.com/u/js-cookie" will trigger despite the branch it was created, as long as it is created by an authorized user (see #13).

By not encoding the comma character as the cookie value in this test branch, it should work in chrome, but fail to set the value in Safari. If the image in the README shows the failure, then it means it does not consider only the "master branch" and the message should be removed.

**EDIT:**

It wasn't necessary to wait for any failure. It indeed changes for any authorized PR, not just commits in the master branch.

image

I precipitated myself on commit https://github.com/js-cookie/js-cookie/commit/04b6544f90c219ca5100889b9a544f2420026839.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -30,8 +30,8 @@ module.exports = function (grunt) {
 				options: {
 					urls: [
 						'http://127.0.0.1:9998/',
-						'http://127.0.0.1:9998/amd.html',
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
Just need to take care by not leaving the status of a failing PR show in the README for a long time.

PR #54, opened 2015-06-24
Fixed: some falsy secure option values modified cookie path

dimac:
Hi,

Found and fixed issue when some falsy values modified cookie path in resulting cookie string due to implicit conversion to string inside Array.join() method.

```javascript
Cookies.set('c', 'v', {secure: false}) === "c=v; path=/false";
Cookies.set('c', 'v', {secure: NaN}) === "c=v; path=/NaN";
Cookies.set('c', 'v', {secure: 0}) === "c=v; path=/0";
```

Expected value for all tests is `"c=v; path=/"`

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -67,7 +67,7 @@
 					attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
 					attributes.path    && '; path=' + attributes.path,
 					attributes.domain  && '; domain=' + attributes.domain,
-					attributes.secure  && '; secure'
+					(attributes.secure  && '; secure') || '' // use '' for falsy values, because [NaN, false, 0].join('') === 'NaNfalse0;
 				].join(''));
 			}
 
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -67,7 +67,7 @@
 					attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
 					attributes.path    && '; path=' + attributes.path,
 					attributes.domain  && '; domain=' + attributes.domain,
-					(attributes.secure  && '; secure') || '' // use '' for falsy values, because [NaN, false, 0].join('') === 'NaNfalse0;
+					(attributes.secure  && '; secure') || ''
 				].join(''));
 			}
 
--- a/test/tests.js
+++ b/test/tests.js
@@ -174,6 +174,33 @@ QUnit.test('API for changing defaults', function (assert) {
 	assert.ok(Cookies.set('c', 'v').match(/path=\//), 'should roll back to the default path');
 });
 
+QUnit.test('falsy secure value', function (assert) {
+	var falsyValues = [
+		false,
+		null,
+		undefined,
+		0,
+		NaN,
+		''
+	];
+
+	assert.expect(falsyValues.length);
+	var expected = 'c=v; path=/';
+
+	falsyValues.forEach(function (secureFalsyValue) {
+		var falsyName;
+
+		if ('' === secureFalsyValue) {
+			falsyName = 'Zero length string';
+		} else {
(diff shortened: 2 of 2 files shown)


FagnerMartinsBrack:
The docs says that secure is:

> A **Boolean** indicating if the cookie transmission requires a secure protocol (https)

It means that what is accepted is a primitive `Boolean`. `secure: NaN` and `secure: 0` are not supported, so the behavior is undefined.

Maybe we could enhance the docs to make it clear that what is supported is the **primitive** type `Boolean`, not the `true` and `false` values of the [`Boolean`](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Boolean) object.

src/js.cookie.js
```diff
@@ -67,7 +67,7 @@
 					attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
 					attributes.path    && '; path=' + attributes.path,
 					attributes.domain  && '; domain=' + attributes.domain,
-					attributes.secure  && '; secure'
+					(attributes.secure  && '; secure') || '' // use '' for falsy values, because [NaN, false, 0].join('') === 'NaNfalse0;
```
FagnerMartinsBrack: Ok, now I get the bug, ignore my last comment.

test/tests.js
```diff
@@ -174,6 +174,33 @@ QUnit.test('API for changing defaults', function (assert) {
...
 
+QUnit.test('falsy secure value', function (assert) {
+	var falsyValues = [
+		false,
+		null,
+		undefined,
+		0,
+		NaN,
+		''
+	];
```
FagnerMartinsBrack: It is not necessary to test for all falsy values, just for the primitive `false`. As stated earlier only the primitive `Boolean` is supported.

src/js.cookie.js
carhartl: Wouldn't it be the same for, say `domain`?

FagnerMartinsBrack: Only if someone declares the domain as `domain: NaN`, `domain: false`, `domain: 0`.

The docs say:

> Define the domain where the cookie is available

Only a `String` is supported.

carhartl: Wouldn't it be the same for, say `domain` (if I happen to pass it to it)?

In that case we would clearly state that it'd be false usage and needed be fixed by them imo.

FagnerMartinsBrack: Expires OK:

```shell
> new Date() && '; expires=' + new Date().toUTCString()
"; expires=Wed, 24 Jun 2015 15:35:51 GMT"
```

Path OK:
```shell
> '/' && '; path=/'
"; path=/"
```

Domain OK:
```shell
> domain.com' && '; domain=domain.com'
"; domain=domain.com"
```

Secure `true` OK:
```shell
> true  && '; secure'
"; secure"
```

Secure `false` FAIL:
```shell
> false  && '; secure'
false
```

Should not print `false`, but ignore it.

FagnerMartinsBrack: > In that case we would clearly state that it'd be false usage and needed be fixed by them imo.

Any behavior that is not documented should not be relied upon. What we could do is to explicitly declare which types each option accepts.

In the `domain` case, the only falsy value that could be passed would be `''`, which will evaluate to an empty domain and be ignored upon `.join('')`:

```shell
> ['' && '; domain='].join('')
""
```

FagnerMartinsBrack:
It appears that your commit doesn't have a github account linked. You can link your github account to the git config using the following commands:

```shell
$ git config --global user.name "My Name In Commits"
$ git config --global user.email my.github.email@example.com
```

@carhartl This PR seems good to me :+1:

dimac:
@FagnerMartinsBrack I've added the address used in commit to my github account, hope this will help.

FagnerMartinsBrack:
I just made some changes for better GZIP compression.

Before:
before

After:
after

PR #60, opened 2015-07-07
decoding of cookie value is done only if necessary

TeamIguana:
decoding of cookie value is done only if necessary


FagnerMartinsBrack:
Hi, thanks for the PR!

Can you please explain the benefit of decoding the cookie only when necessary?

* If this is a performance issue, in which situation it hangs?
* If this is fixing something, can you please provide a test case in the `/test` folder to prevent regression?

TeamIguana:
Hi,
we encountered a problem when third part javascript has set a cookie with a value that can't be decodified (for example if the cookie value is '...%A2...' that is a not properly encoded text).

With this modification it is avoided to decode cookies that are not requested.

In general it can be positive for performance if you have a lot of analytics or marketing javascripts because they generate hundreds of cookies.

The problem of a cookie value with a not properly encoded text could be solved by catching the exception raised, avoiding that an error on a single cookie could break every javascript in a page, but this will be another pull request.

FagnerMartinsBrack:
> we encountered a problem when third part javascript has set a cookie with a value that can't be decodified (for example if the cookie value is '...%A2...' that is a not properly encoded text).

That's why converters were created. js-cookie is supposed to follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle), encoding only what is necessary but allowing any cookie to exist. If a third party cookie cannot be decoded by js-cookie, then you can use a [converter](https://github.com/js-cookie/js-cookie#converter) to create a new instance with a specific decoding configuration.

> In general it can be positive for performance if you have a lot of analytics or marketing javascripts because they generate hundreds of cookies.

The spec imposes [some limits](http://tools.ietf.org/html/rfc6265#section-6.1) of how many cookies should exist in the browser. One should not have hundreds of cookies in the same domain because the limit is 50.

Usually we try to be conservative regarding performance and focus only in cases where there's an evidence that the app is hanging. One of the features of `js-cookie` is it's lightweightness, even with a lot of features we managed to cut it down to ~800 bytes gzipped. Any change that adds more code should take care not to exceed that.

> avoiding that an error on a single cookie could break every javascript in a page

That's important to address. js-cookie should fail only for a single cookie if it does not conforms with the desired encoding. If it is currently failing because of the existence of another cookie in the page, then it is indeed a bug and should be addressed.

FagnerMartinsBrack:
> The problem of a cookie value with a not properly encoded text could be solved by catching the exception raised

Do you want to send a PR on this one?

FagnerMartinsBrack:
I committed the fix for values encoded using `escape` breaking the use of js-cookie in the page (https://github.com/js-cookie/js-cookie/commit/e36ddd2e3268627328a5a198384001533af2a67a). That was a regression from jquery-cookie (more information in the related PR), thanks very much for reporting it!

Regarding the intent of this PR, I checked out your branch to see how many bytes this modification added.

Before (https://github.com/TeamIguana/js-cookie/commit/76932e33d9b4a50789c2edb0a1aa9657fde9b829):
image

After (https://github.com/TeamIguana/js-cookie/commit/87b7057238a4ab7e3b7d84417dfa0cd5a0f79210):
image

As you can see, executing the decoding operation for all cookies helps to cut `24 bytes` gzipped without any evidence of performance bottleneck.

Issue #55, opened 2015-06-29
malformed URI sequence


flanderboy:
Hello,
I used the script but I get this error

malformed URI sequence...
cookie = converter && converter(cookie, name) || cookie.replace(rdecode, decodeURIComponent);

I'm writing a cookie in this way
Cookies.set('screech', 'hide', {expires: 1});

Stefano

FagnerMartinsBrack:
That error occurs while decoding the cookies.

Which version are you using?
Can you please post the contents of your `document.cookie`?

flanderboy:
Hello,
I do not decoded the cookie. I checked only if there is.

if (!Cookies.get('screech')) {

}

FagnerMartinsBrack:
> I do not decoded the cookie.

The decode is done automatically when you call `Cookies.get()`. The error probably represents another cookie with an invalid format.

I need the following information to understand which cookie it is:

1. Open the developer console
2. Type `document.cookie`
3. Post the returning data.

Example below:

![example](https://cloud.githubusercontent.com/assets/835857/8411129/a7b81cbc-1e58-11e5-96bb-efa337383c2d.gif)

flanderboy:
"PHPSESSID=4h2skf7q3bh5ucvguni41918k0; screech=hide"

FagnerMartinsBrack:
Are you sure what you posted is the content of `document.cookie` when the error happens?

flanderboy:
yes

FagnerMartinsBrack:
I tested the following code with [the master version pf js.cookie.js](https://github.com/js-cookie/js-cookie/blob/76932e33d9b4a50789c2edb0a1aa9657fde9b829/src/js.cookie.js) and everything is fine, there's no error:

```html
<!DOCTYPE html>
<html>
	<head>
		<script src="js.cookie.js"></script>
	<head>
	<body>
		<script>
			window.onload = function() {
				document.cookie = "PHPSESSID=4h2skf7q3bh5ucvguni41918k0; screech=hide";
				Cookies.set( "screech", "hide", {
					expires: 1
				});
				if ( !Cookies.get( "screech" ) ) {
					alert( "no cookie!" );
				}
			};
		</script>
	</body>
<html>
```

FagnerMartinsBrack:
[...]

PR #62, opened 2015-07-10
Should not prevent a cookie to be retrieved if there is another malformed one

FagnerMartinsBrack:
If there is a malformed cookie that cannot be decoded using the default decoding mechanism, the malformed exception should be ignored and fail to read that single cookie. It should not prevent an unrelated cookie to be retrieved.

This seems to be related to https://github.com/carhartl/jquery-cookie/commit/81dd6a1596bdcd8e6c4008fcfc6d42825a7c4a2f. What happened is that the original code tested for a single percent character, while in this case the encoding is in a valid percent-encoding format that fails to be decoded by `decodeURIComponent` spec.

Brought up thanks to @TeamIguana, in his [comment](https://github.com/js-cookie/js-cookie/pull/60#issuecomment-119565763).

It might be related to the following report: https://github.com/js-cookie/js-cookie/issues/55

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -93,22 +93,24 @@
 					cookie = cookie.slice(1, -1);
 				}
 
-				cookie = converter && converter(cookie, name) || cookie.replace(rdecode, decodeURIComponent);
+				try {
+					cookie = converter && converter(cookie, name) || cookie.replace(rdecode, decodeURIComponent);
 
-				if (this.json) {
-					try {
-						cookie = JSON.parse(cookie);
-					} catch (e) {}
-				}
+					if (this.json) {
+						try {
+							cookie = JSON.parse(cookie);
+						} catch (e) {}
+					}
 
-				if (key === name) {
-					result = cookie;
-					break;
-				}
+					if (key === name) {
+						result = cookie;
+						break;
+					}
 
-				if (!key) {
-					result[name] = cookie;
-				}
+					if (!key) {
+						result[name] = cookie;
+					}
+				} catch (e) {}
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
This adds 4 bytes gzipped

sem titulo

FagnerMartinsBrack:
Here is the real world issue:

It is pretty common to have arbitrary cookies created using `escape`. If one have an unrelated cookie that was encoded using `escape` and can only be decoded using `unescape`, it should not throw a `Malformed URI` exception using the default decoding mechanism. it should just ignore it.

```javascript
unescape( "%A2" ); // ¢
decodeURIComponent( "%A2" ); // Uncaught URIError: URI malformed
```

Since the `%A2` cookie will be ignored, the solution to be able to read it is to use a converter to handle each specific malformed cookie, something like:

```javascript
var SafeCookies = Cookies.withConverter(function( value, name ) {
  try {
    // Just check if the default decoding will fail for this value.
    // If it's not going to fail, just return undefined and use the default decoding mechanism.
    decodeURIComponent( value );
  } catch(e) {
    // If it fails, we just assume this was encoded using "escape".
    // You can create more complex rules to account for all possible
    // encoding formats and unit test it.
    return unescape( value );
  }
});
SafeCookies.get( 'escaped-cookie' ); // no failure
SafeCookies.get( 'default-encoded-cookie' ); // no failure
```

For more information, read the [converter docs](https://github.com/js-cookie/js-cookie/tree/4a67cb3bda1d7bc344d7339ae4efc38d890eb47c#converter).

PR #67, opened 2015-07-24
Fix Cookie setup as window global

carhartl:
No longer pass jQuery to factory().


FagnerMartinsBrack:
:+1:

FagnerMartinsBrack:
FWIW, this removed 8 bytes gzipped:

image

PR #68, opened 2015-07-28
The JSON-js polyfill note in the README should not include IE8

yanxyz:
IE8 supports JSON


FagnerMartinsBrack:
IE8 requires the document to be in IE8+ [standards mode](http://msdn.microsoft.com/en-us/library/cc288325%28VS.85%29.aspx) for `JSON` to work: http://caniuse.com/#feat=json

Issue #70, opened 2015-07-31
Optional encoder function, in addition to decoder (converter)


dezlov:
This will allow for ultimate customization and compatibility with other cookie encoding functions.

It could be added as an extension to `Cookies.withConverter()` to allow specifying an encoding function too.

For example, the incompatibility with PHP cannot be resolved because this JS library will always store "+" signs in plain text, while PHP will always convert them to spaces when read.

In PHP, both `$_COOKIE` and `filter_input(INPUT_COOKIE)` will return already decoded value, where all "+" signs will be converted to spaces. There is a `setrawcookie`, but no `getrawcookie` function, so no way around it in PHP (without having to manually double encode and decode in both languages).

dezlov:
Related ticket: https://github.com/carhartl/jquery-cookie/issues/13

It only addresses one-way problem (PHP to JS), but not the reverse (JS to PHP).

FagnerMartinsBrack:
Are you sure there is no way to do it in PHP? The `+` sign is a valid character in the RFC, I know .net has some method to workaround it, PHP should have it too.

Did you consider reporting to PHP itself as a bug report due to the fact it handles the RFC incorrectly?

dezlov:
Yes, I'm pretty sure.

PHP reads cookie string and replaces all '+' characters with spaces, for historical reasons (as mentioned in the docs).

I guess one can parse HTTP headers manually and extract all cookie information manually, but the native PHP methods for accessing cookies via `$_COOKIE` and `filter_input(INPUT_COOKIE)` have this '+' handling built-in and not adjustable.

To make this library compatible with PHP cookie encoding/decoding: always encode `+` as `%2B`. The diff below demonstrates how to apply it in code.

```diff
59c59,60
< 				value = value.replace(/%(23|24|26|2B|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g, decodeURIComponent);
---
> 				// do not decode '2B' back to '+' for PHP compatability!
> 				value = value.replace(/%(23|24|26|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g, decodeURIComponent);
62c63,64
< 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
---
> 				// do not decode '2B' back to '+' for PHP compatability!
> 				key = key.replace(/%(23|24|26|5E|60|7C)/g, decodeURIComponent);
[...]

PR #71, opened 2015-08-01
Write converter

FagnerMartinsBrack:
When calling `Cookies.withConverter`, you can pass either a function to
use as the read converter, or an Object Literal with the properties
`read` and `write`.

image


FagnerMartinsBrack:
image

FagnerMartinsBrack:
(@dezlov, it seems that your comment was removed on rebase, so I will replicate here:)

@dezlov:
I didn't have chance to verify that converters above actually work with PHP, but I've noticed few other problems.

1. ~~Typo in `Cookies.withConverter(...)`, it is called with 2 x `write` properties.~~ (fixed in latest version)
2. Usage of `setrawcookie()` is missing as documented in https://github.com/js-cookie/js-cookie/issues/70#issuecomment-126965203, it's quite important:

----

**In PHP**, use `setrawcookie()` instead of `setcookie()`:

```php
setrawcookie($name, rawurlencode($value));
```

FagnerMartinsBrack:
> I didn't have chance to verify that converters above actually work with PHP

I didn't tested it either. It is just a reference documentation to workaround server-side incompatibilities based in PHP docs and Tomcat's source code.

> In PHP, use setrawcookie() instead of setcookie():

I will add that

FagnerMartinsBrack:
@dezlov, added.

Issue #57, opened 2015-07-01
Implement an automatic release process


FagnerMartinsBrack:
Currently, everything is done manually, I just [document the steps](https://github.com/js-cookie/js-cookie/tree/76932e33d9b4a50789c2edb0a1aa9657fde9b829#manual-release-steps).

Here are a few things that should be fixed after a proper release process:
- <strike>[Minification for jsDelivr](https://github.com/jsdelivr/jsdelivr/pull/5507#issuecomment-116729066)</strike> (jsDelivr has [on-demand minification now](https://github.com/js-cookie/js-cookie/pull/371#issuecomment-339047451))
- <strike>[Document the CDN file](https://github.com/js-cookie/js-cookie/pull/46#issuecomment-113061185)</strike> (We're documenting [jsDelivr](https://github.com/js-cookie/js-cookie/pull/371#issuecomment-340131636) for now)
- [Cleanup the README by removing the manual steps](https://github.com/js-cookie/js-cookie/tree/76932e33d9b4a50789c2edb0a1aa9657fde9b829#manual-release-steps)
- [Play nice with Django by releasing the minified file on NPM](https://github.com/js-cookie/js-cookie/issues/66)
- [Request to add the minified version in bower](https://github.com/js-cookie/js-cookie/issues/73)
- [Create a `js.cookie.debug.js` file for sanity checks regarding wrong domain](https://github.com/js-cookie/js-cookie/issues/86) (Maybe use `drop_console` option from UglifyJS?)
- [Create a `js.cookie.debug.js` file for sanity checks regarding maximum cookie size](https://github.com/js-cookie/js-cookie/issues/88)

Pull request: #82

carhartl:
Just saw this, maybe worth it to take a closer look: https://medium.com/@webprolific/using-release-it-60b96515c073

FagnerMartinsBrack:
[Added jsDelivr CDN to the README](https://github.com/js-cookie/js-cookie/pull/371#issuecomment-340131636) because now they support on-demand minification.
The bullet points related to the CDN from the PR description have been removed because of that.

FagnerMartinsBrack:
Update: Haven't had time to continue on [js-cookie-release](https://github.com/js-cookie/js-cookie-release), be advised this may take a long time to be implemented. Automate a release with proper tests to make sure all the Git commits always work fine seem to be harder than what I thought.

FagnerMartinsBrack:
[...]

Issue #35, opened 2015-05-04
Why decodeURIComponent when writing


andraaspar:
What is the purpose of [this line](https://github.com/js-cookie/js-cookie/blob/master/src/js.cookie.js#L59):

```
value = value.replace(/%(23|24|26|2B|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g, decodeURIComponent);
```

When writing a cookie? This line reverts some of the characters converted by encodeURIComponent, right? Is that only to make the value more readable, or does it serve any other purpose?

FagnerMartinsBrack:
Those characters are allowed in a cookie value, so there's no need to encode it. js-cookie encodes only the characters that are not allowed in the cookie value:

https://github.com/js-cookie/js-cookie/tree/1a0d1ed13c106382d2cfe1186a73e8091e4ac349#encoding

> This project is RFC 6265 compliant. All special characters that are not allowed in the cookie-name or cookie-value are encoded with each one's UTF-8 Hex equivalent using percent-encoding.
The only character in cookie-name or cookie-value that is allowed and still encoded is the percent % character, it is escaped in order to interpret percent input as literal.

Check here for context: https://github.com/js-cookie/js-cookie/pull/20#issuecomment-94453407

and here for the spec: http://tools.ietf.org/html/rfc6265#section-4.1.1

> Is that only to make the value more readable, or does it serve any other purpose?

It just to minimize the amount of encoded characters to leverage what browsers support and to save bytes. There is a limit of 4096 bytes per cookie, if we encode valid characters like `+`, then we are transmitting 3 unnecessary bytes to the server (`%2B`) for a character that is globally supported.

andraaspar:
I see, a very nice feature. Thanks for the explanation!

PR #83, opened 2015-09-02
add files to bower ignore

timse:
this removes clutter when installing `js-cookie` as a bower component

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,5 +3,10 @@
   "main": [
     "src/js.cookie.js"
   ],
-  "ignore": []
+  "ignore": [
+    "travis.sh",
+    "test",
+    "Gruntfile.js",
+    "package.json"
+  ]
 }
--- a/bower.json
+++ b/bower.json
@@ -7,6 +7,11 @@
     "travis.sh",
     "test",
     "Gruntfile.js",
-    "package.json"
+    "package.json",
+    ".gitignore",
+    ".jshintignore",
+    ".jshintrc",
+    ".tm_properties",
+    ".travis.yml"
   ]
 }


bower.json
```diff
@@ -3,5 +3,10 @@
   "main": [
     "src/js.cookie.js"
   ],
-  "ignore": []
+  "ignore": [
+    "travis.sh",
+    "test",
```
FagnerMartinsBrack: We need this folder to be able to [integrate the encoding tests for server-side frameworks](https://github.com/js-cookie/js-cookie/blob/v2.0.3/CONTRIBUTING.md#integration-with-server-side).

timse:
I'm not sure i understand this. Why would you want to integrate a third parties tests?

bower.json
timse: I'm not sure i understand this. Why would you want to integrate a third parties tests?

FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/commit/e95fde0c6205f992dbab7a7d189726cfd2e38213, most of those files are needed in order to run the build after the files are retrieved using bower.

[Here](https://github.com/js-cookie/java-cookie/blob/v0.0.2/Gruntfile.js#L8) is the example of a project that uses it.

> I'm not sure i understand this. Why would you want to integrate a third parties tests?

To prevent having to duplicate each character encoding/decoding test in a different server-side language.

FagnerMartinsBrack:
js-cookie encodes [only the characters that are not allowed in cookies](https://github.com/js-cookie/js-cookie/tree/v2.0.3#encoding), as specified in the RFC 6265. See [here](https://github.com/js-cookie/js-cookie/issues/35) the details. The integration test hook is exposed to be able to ensure consistency between server-side implementations with the default js-cookie decoding/encoding mechanism, like the Java Cookie project.

timse:
> To prevent having to duplicate each character encoding/decoding test in a different server-side language.
integrating the whole javascript toolchain to save up on a test doesnt look too like i want to do that. But i guess you can do whatever you feel like with your package :)

FagnerMartinsBrack:
If you are using bower to hold only production-ready components you are doing something different from what it was built for, see https://github.com/bower/bower/issues/368#issuecomment-15916138

> The intention has never been for Bower to provide production-ready components. If someone wants to register a package that includes minified files to reference in your script/link elements or something, then that's fine. But that's not Bower's concern. Publishing minified files is something that might be more common if we decide to include an npm-like publish model.

timse:
i did read the comment the first time you linked to it. they talk about not providing minified code which i totally agree with, as it would be just more clutter.

they are not talking about adding the whole test suit so you can do weird things with it. 

i'd rather add extract your test case as a bower component and require it in here as a dev dependency but then again its not my package and you are free to do whatever you want with it. I just very much disagree.

FagnerMartinsBrack:
@timse I understand your concerns. One more reason to keep the test folder is to make it possible to run `grunt default` in a custom build environment that consumes js-cookie package, in order to be able to generate the minified file for production.

[We have plans](https://github.com/js-cookie/js-cookie/issues/57) to provide the minified file without having to run the build, mostly for npm. Maybe after that it would make sense moving the encoding tests to a separate package and clear all the clutter for those downloading the full package once and for all.

I will leave this out for @carhartl to decide what's the best way to go for now.

carhartl:
I'm not familiar with how the server-side testing is being setup, but I would assume we could pull in a git tag instead of a bower package to set it up?

If so, I'd prefer not to clutter the bower package, i.e. +1 for this PR.

Regarding the minification it provides at the moment, imo we should not impose Grunt on the users of the package, by now Gulp seems even more popular. That said, the task may be useless too many and declare a dependency that isn't satisfied in the first place. Furthermore, developers likely have set up their own build tool chain, an argument that was often made by us in the past against including a minified version in the git repository. 

If minification etc. is a common expectation for bower packages to provide, then I'll change my mind on that one :)

FagnerMartinsBrack:
> I'm not familiar with how the server-side testing is being setup

It integrates a front-end build with the server-side build routine, executing bower eventually to load js-cookie files.

> I would assume we could pull in a git tag instead of a bower package to set it up?

With [submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules), probably yes.

Very good point about grunt/gulp. Different from [the last removal](https://github.com/carhartl/jquery-cookie/pull/205), now we have a proper justification for removing this.

bower.json
FagnerMartinsBrack: @timse As from [this comment](https://github.com/js-cookie/js-cookie/pull/83#issuecomment-143968146), doesn't it makes sense removing everything then, including the dot files? Like `.gitignore`, `.jshintignore`, `.jshintrc`, `.travis.yml` and `.tm_properties`?

It seems that the only things that matters are `README.md`, `MIT-LICENSE.txt`, `CONTRIBUTING.md` (is it needed?), and `src/js.cookie.js`, right?

FagnerMartinsBrack:
@timse Are you still willing to continue on this?

bower.json
timse: i guess that would be finde with me

FagnerMartinsBrack: Thanks for the feedback, do you want to update the PR?

timse: Sure i will as soon as im back at my computer :)

FagnerMartinsBrack: :+1:

FagnerMartinsBrack:
Thanks man :+1:

timse:
thanks! :) :+1:

PR #85, opened 2015-09-11
Update README to include package version number in npm and bower

AlexVPopov:


FagnerMartinsBrack:
It might be interesting to highlight that there is a package in npm and bower. I would say that in the "Installation" section the "js-cookie supports npm and Bower under the name js-cookie" could be removed.

@carhartl

carhartl:
Is there a possibility that the bower and npm versions would differ at any time? At least when we were releasing a new version we would make sure to have both packages up-to-date... so in that respect it seems redundant to include versions for those, no? And the latest release to lookup would always be in the repositories releases. 

Npm has pages for packages, which I'd like to add a direct link to, I think Bower does not provide such a page?

carhartl:
It doesn't hurt of course to add these versions for people needing to take a quick glance.

carhartl:
Regarding that sentence in the readme, we could just add instructions "Install with npm/install with bower" which would then make package manager support implicit, plus people like to quickly copy&paste the commands...

@FagnerMartinsBrack p

FagnerMartinsBrack:
> Is there a possibility that the bower and npm versions would differ at any time? 

It depends. Bower is unopinionated, so one can use either `master`, a specific commit `a5464f654` or a semver pattern `1.0.x`. Most people usually load the latest release tag. Bower doesn't have an actual registry, it just acts as an adapter to github repositories, making requests to github code when a package is requested.

> so in that respect it seems redundant to include versions for those, no?

Agree, that's redundant. The only reason for having the tags then would be to better highlight that it can be used in both.

> I think Bower does not provide such a page?

Yeah, the only thing it has is the link to the search page: http://bower.io/search/?q=js-cookie. Since bower doesn't have a registry it assumes the "package homepage" is the github repo when you click in `js-cookie` there.

I can see the value of showing the latest version for npm, but I don't see a reason for bower. If it's published in npm it means there is a release. If there is a release, bower can pick it up, unless we explicitly declare in the badge (which is one more thing to maintain).

As you can see below, if we don't declare a version in the bower badge, it doesn't know which version to pick.

[npm version](http://badge.fury.io/js/js-cookie) [Bower version](http://badge.fury.io/bo/js-cookie)

FagnerMartinsBrack:
@AlexVPopov It seems we will need to document how to install using npm and bower, do you want to take a look on this?

rogerangers:
Hi!

About js-cookie, about write spaces:

Cookies.set('test', '               1             2            ');

Cookies.get('test')  = > return  ' 1 2 '

rogerangers:
Hi!

About js-cookie, about write spaces:

var space = ' ';

Cookies.set('test',space+space+space+'1'+space+space+space+space+'2'+space+space+space);

Cookies.get('test')  = > return  ' 1 2 '

FagnerMartinsBrack:
@rogerangers please open a new issue.

FagnerMartinsBrack:
Closing due to the lack of feedback from the OP.

rogerangers:
Hi!

About js-cookie, about write spaces:

var space = ' '; //one space

Cookies.set('test',space+space+space+'1'+space+space+space+space+'2'+space+space+space);

Cookies.get('test') = > return ' 1 2 '

PR #102, opened 2015-10-09
Optimize the `var` declarations.

cdp1337:
This should result in slightly better compression ratios, (and less complaining from YUI).

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -6,13 +6,14 @@
  * Released under the MIT license
  */
 (function (factory) {
+	var _OldCookies, api;
 	if (typeof define === 'function' && define.amd) {
 		define(factory);
 	} else if (typeof exports === 'object') {
 		module.exports = factory();
 	} else {
-		var _OldCookies = window.Cookies;
-		var api = window.Cookies = factory();
+		_OldCookies = window.Cookies;
+		api = window.Cookies = factory();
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
Can you show the difference in the GZip compression ratio? We have a task that does it called `grunt compare-size`, see [the docs here](https://github.com/rwaldron/grunt-compare-size) and the [task config](https://github.com/js-cookie/js-cookie/blob/5afa5d967d8f6b5d2a498926ae3e98e9a1c4857d/Gruntfile.js#L88).

cdp1337:
Honestly I'm not sure, and grunt isn't in the debian repo...

FagnerMartinsBrack:
By the way, you commited `js.cookie.min.js`, that's not necessary since we don't keep minified files in the source repo (we have plans to create a [release process that fixes that and other issues](https://github.com/js-cookie/js-cookie/issues/57)).

FagnerMartinsBrack:
I tested this:

```
$ rmdir build /s /q
$ git reset --hard 5afa5d96
$ grunt && grunt compare_size:save:previous_commit
$ git reset --hard a6b08495
$ rmdir build /s /q
$ grunt && grunt compare_size
```

Result:

image

It does have a change in the raw size (`-14`) , but the gzipped size gains an increase of +2 bytes minified.

PR #110, opened 2015-11-17
Fix problem when minifying with Uglify2

pjeweb:
Code did not work because basically `[1, undefined, 2].join('')` gets minified to `1 + undefined + 2` in Uglify2 with compress.evaluate option set to true.
Also make code better understandable :)

Tested with https://skalman.github.io/UglifyJS-online/:
```javascript
function a() {
  return (document.cookie = [
    key, '=', value,
    attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
    attributes.path    && '; path=' + attributes.path,
    attributes.domain  && '; domain=' + attributes.domain,
    attributes.secure ? '; secure' : ''
  ].join(''));
}
```
```javascript
function a(){return document.cookie=key+"="+value+(attributes.expires&&"; expires="+attributes.expires.toUTCString())+(attributes.path&&"; path="+attributes.path)+(attributes.domain&&"; domain="+attributes.domain)+(attributes.secure?"; secure":"")}
```

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -66,7 +66,7 @@
 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
 				key = key.replace(/[\(\)]/g, escape);
 
-				var cookieString = key, '=', value;
+				var cookieString = key + '=' + value;
 				if (attributes.expires) {
 					cookieString += '; expires=' + attributes.expires.toUTCString(); // use expires attribute, max-age is not supported by IE
 				}
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -66,13 +66,22 @@
 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
 				key = key.replace(/[\(\)]/g, escape);
 
-				return (document.cookie = [
-					key, '=', value,
-					attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
-					attributes.path    && '; path=' + attributes.path,
-					attributes.domain  && '; domain=' + attributes.domain,
-					attributes.secure ? '; secure' : ''
-				].join(''));
+				var cookieString = key, '=', value;
+				if (attributes.expires) {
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
I will take a look on it, but it seems to be a problem with Uglify2.

pjeweb:
Yes it's UglifyJS's problem but still this code relies on a not easily understandable feature of JS.
Another option would be to use the <code>foo ? 'bar' : '';</code> syntax as it is used for <code>attributes.secure</code> already, preventing <code>undefined</code> in the array. This way you don't need to change as much. It stays compact while being consistent :)

FagnerMartinsBrack:
> this code relies on a not easily understandable feature of JS

That's a trade-off that was made in order to maintain the "**~800 bytes** gzipped" bullet point documented in the ["Features" section of the README](https://github.com/js-cookie/js-cookie/tree/0fe470c76d254fa413c56bd83e77fcb0a6d5f29b#javascript-cookie--). Right now we have 874 bytes gzipped as from #71.

I understand the legibility argument and I agree with that. The point is that we have currently 100% test coverage and use TDD extensively, so it is unlikely that a bug arises from now on because of those legibility problems.

Feel free to throw a `grunt compare-size` comparison. If it cut a single byte and there is no external implications, then I will be willing to merge it :+1:

pjeweb:
There is apparently a pull request for that issue in Uglify from 2013 (https://github.com/mishoo/UglifyJS2/issues/301). Hope it gets fixed there. When I have the time I will look into <code>grunt compare-size</code> though.

FagnerMartinsBrack:
In the mean time you can use the minified version from the [Releases page](https://github.com/js-cookie/js-cookie/releases) or build your own minified file using:

```
$ grunt
```

pjeweb:
Well we have a running build process (using gulp, babel, browserify) and integrating another build process into that is not really worth it.
But as a comment on the issue (https://github.com/mishoo/UglifyJS2/pull/301#issuecomment-158124405) on Uglify noted it was due to an <code>unsafe</code> option that I somehow enabled in all my tests :)

pjeweb:
Btw thanks for responding fast and this project :+1:

FagnerMartinsBrack:
You're welcome. Good hacking :+1:

PR #119, opened 2015-12-30
Create js-cookie.min.js

SamsamBabadi:


FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/issues/120#issuecomment-168006064

PR #123, opened 2016-01-08
Adding remove all

luanpotter:
This is an interesting feature that I tend to use in several places in most of my projects.  
I believe it would be a neat addition to the api.  
I'm open to suggestions, of course :)

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -136,6 +136,12 @@
 			}));
 		};
 
+		api.removeAll = function (attributes) {
+			Object.keys(api.get()).forEach(function (cookie) {
+				api.remove(cookie, attributes);
+			});
+		};
+
 		api.withConverter = init;
 
 		return api;
--- a/test/tests.js
+++ b/test/tests.js
@@ -217,6 +217,14 @@ QUnit.test('deletion', function (assert) {
 	assert.strictEqual(document.cookie, '', 'should delete the cookie');
 });
 
+QUnit.test('delete all', function (assert) {
+	assert.expect(1);
+	Cookies.set('a', 'v1');
+	Cookies.set('b', 'v2');
+	Cookies.removeAll();
+	assert.strictEqual(document.cookie, '', 'should delete all the cookies');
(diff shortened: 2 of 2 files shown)


FagnerMartinsBrack:
We refrain from adding additional features to support unless there is enough evidence to justify the cost. Is there anything that should be done in `js-cookie` side that prevents you from extending `js-cookie` and doing something like this locally?

```javascript
Cookies.removeAll = function (attributes) {
 	Object.keys(Cookies.get()).forEach(function (cookie) {
 		Cookies.remove(cookie, attributes);
 	});
 };
```

`js-cookie` was created to remove the complexity of handling cookies and allow code like that to be created in the developer side. I see no reason for adding a new method that do not have the same benefit of abstracting `document.cookie = String`

This would be an amazing [one-line](https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328)-like module though, and I would certainly use it to remove all cookies accessible in the current page.

@carhartl

luanpotter:
No, it works with the extension -- in fact, I do that already in my projects.

I agree it would be overall clutter. And you gave a nice idea. I will see about making this one-line separate npm module, and so those who want can add this functionality with negligible effort, while not starting to attach little bits sparsely hanging around in the lib.

Thanks for the quick feedback!

FagnerMartinsBrack:
When you create it, please link here for other ppl to find out. Maybe we can add something in the wiki linking to the module as a plugin? A self-answered question in StackOverflow would be amazing for SEO and visiblity too if the question constructs the url as `remove-all-cookies-using-javascript` ;)

luanpotter:
Done: https://github.com/luanpotter/js-cookie-remove-all

FagnerMartinsBrack:
:+1:

FagnerMartinsBrack:
I have contributed with a PR for your repo, see https://github.com/luanpotter/js-cookie-remove-all/pull/1
Since we are delegating this to `js-cookie-remove-all`, I will assume the issue as solved.

PR #131, opened 2016-01-30
Update SERVER_SIDE.md

ThiagoDosSantos:
Say Tomcat 8 to be compliant with RFC6265

### DIFF
--- a/SERVER_SIDE.md
+++ b/SERVER_SIDE.md
@@ -71,6 +71,15 @@ var TomcatCookies = Cookies.withConverter({
 
 Alternatively, you can check the [Java Cookie](https://github.com/js-cookie/java-cookie) project, which integrates nicely with JavaScript Cookie.
 
+## Tomcat 8
+
+Since Tomat 8.0.15, there is a simple way to [say tomcat to be compliant with the RFC 6265](https://tomcat.apache.org/tomcat-8.0-doc/config/cookie-processor.html). It comes down to change your conf/context.xml file, adding the new CookieProcessor element nested inside the Context element. It would be like this:
+
+```xml
+<Context> <CookieProcessor className="org.apache.tomcat.util.http.Rfc6265CookieProcessor"/> </context>
+```
+And you're all done.
+
 ## JBoss 7.1.1
 
 It seems that the servlet implementation of JBoss 7.1.1 [does not read some characters correctly](https://github.com/js-cookie/js-cookie/issues/70#issuecomment-148944674), even though they are allowed as per [RFC 6265](https://tools.ietf.org/html/rfc6265#section-4.1.1). To fix this you need to write a custom converter to send those characters correctly.


SERVER_SIDE.md
```diff
@@ -71,6 +71,15 @@ var TomcatCookies = Cookies.withConverter({
 
 Alternatively, you can check the [Java Cookie](https://github.com/js-cookie/java-cookie) project, which integrates nicely with JavaScript Cookie.
 
+## Tomcat 8
+
+Since Tomat 8.0.15, there is a simple way to [say tomcat to be compliant with the RFC 6265](https://tomcat.apache.org/tomcat-8.0-doc/config/cookie-processor.html). It comes down to change your conf/context.xml file, adding the new CookieProcessor element nested inside the Context element. It would be like this:
```
FagnerMartinsBrack: Maybe the mid part would fit better if expressed like "... Since Tomcat 8.0.15, it is possible to configure RFC 6265 compliance by changing your \[conf/context.xml](http://link-to-some-revision-with-the-file-to-be-used-as-example?) file and adding the new \[CookieProcessor](https://tomcat.apache.org/tomcat-8.0-doc/config/cookie-processor.html) element ..."

SERVER_SIDE.md
```diff
@@ -71,6 +71,15 @@ var TomcatCookies = Cookies.withConverter({
 
 Alternatively, you can check the [Java Cookie](https://github.com/js-cookie/java-cookie) project, which integrates nicely with JavaScript Cookie.
 
+## Tomcat 8
```
FagnerMartinsBrack: I wonder if we should use a new topic for a new Tomcat version. What if we change everything to a single "Tomcat" topic and create subtopics for 7.x and 8.x so we don't start cluttering the "root level" of the servers file. What do you think, does that makes sense?

ThiagoDosSantos: Yes man, It does.

SERVER_SIDE.md
ThiagoDosSantos: Hum, maybe it would add unnecessary complexity to something that is really simple to get done.

FagnerMartinsBrack: What do you mean by "unnecessary complexity"?

ThiagoDosSantos: Complexity is not the word, my bad. The part I think is unnecessary is a link to http://link-to-some-revision-with-the-file-to-be-used-as-example. As the configuration is so simple, I think it should be as the original way. But if you want, I can do it.

SERVER_SIDE.md
ThiagoDosSantos: How can I do that? I'm not so good with git.

SERVER_SIDE.md
FagnerMartinsBrack: Sure, I agree the link is unnecessary. What about my suggestion of how it is structured, does it explain better?

"... Since Tomcat 8.0.15, it is possible to configure RFC 6265 compliance by changing your conf/context.xml file and adding the new \[CookieProcessor](https://tomcat.apache.org/tomcat-8.0-doc/config/cookie-processor.html) element ..."

SERVER_SIDE.md
FagnerMartinsBrack: You can do something called **git amend**. Git has 2 states before commit, the modified state (when you modify a file), and the staged state (when you have staged your changes to commit). Assuming you are starting everything from scratch you will need to:

1. [Install Git for your OS](https://git-scm.com/) and make the `git` command available in the command line.
2. Choose a directory to setup the repository and run the command `git clone https://address-of-your-forked-repository.git`. It will create the `js-cookie` directory with the project structure there.
3. Edit the SERVER_SIDE file with the desired changes
4. Add the file to the staging area after you modified it using `git add SERVER_SIDE.md`
5. Run the command to amend the changes to the latest commit: `git commit --amend`

Then, after you have commited you just push the changes to your fork using the command `git push <remote> <source-branch>:<destination-branch>`. In this case it should be something like `git push origin patch-1:patch-1`

After you pushed, this Pull Request will update automatically.

FagnerMartinsBrack:
@ThiagoDosSantos Are you still tracking this, need any help?

PR #159, opened 2016-03-22
Make sure js-cookie does not break Node JS applications

ericntd:
I've been using js-cookie for my React application and it's been wonderful.

However, I now need to use Express JS to do server side rendering of my application.
The use of `document` is breaking such process.

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -74,7 +74,9 @@
 						attributes.domain  && '; domain=' + attributes.domain,
 						attributes.secure ? '; secure' : ''
 					].join(''));
-				} else return '';
+				} else {
+					return '';
+				}
 			}
 
 			// Read
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -65,14 +65,16 @@
 				key = encodeURIComponent(String(key));
 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
 				key = key.replace(/[\(\)]/g, escape);
-
-				return (document.cookie = [
-					key, '=', value,
-					attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
-					attributes.path    && '; path=' + attributes.path,
-					attributes.domain  && '; domain=' + attributes.domain,
-					attributes.secure ? '; secure' : ''
-				].join(''));
+				
+				if (typeof document !== 'undefined')
+					return (document.cookie = [
+						key, '=', value,
+						attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
+						attributes.path    && '; path=' + attributes.path,
+						attributes.domain  && '; domain=' + attributes.domain,
+						attributes.secure ? '; secure' : ''
+					].join(''));
+				else return '';
 			}
 
 			// Read
@@ -84,7 +86,7 @@
 			// To prevent the for loop in the first place assign an empty array
 			// in case there are no cookies at all. Also prevents odd result when
 			// calling "get()"
-			var cookies = document.cookie ? document.cookie.split('; ') : [];
+			var cookies = (typeof document !== 'undefined' && document.cookie) ? document.cookie.split('; ') : [];
 			var rdecode = /(%[0-9A-Z]{2})+/g;
 			var i = 0;
 
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -66,7 +66,7 @@
 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
 				key = key.replace(/[\(\)]/g, escape);
 				
-				if (typeof document !== 'undefined')
+				if (typeof document !== 'undefined') {
 					return (document.cookie = [
 						key, '=', value,
 						attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
@@ -74,7 +74,7 @@
 						attributes.domain  && '; domain=' + attributes.domain,
 						attributes.secure ? '; secure' : ''
 					].join(''));
-				else return '';
+				} else return '';
 			}
 
 			// Read
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -1,10 +1,10 @@
 /*!
- * JavaScript Cookie v2.1.0
- * https://github.com/js-cookie/js-cookie
- *
- * Copyright 2006, 2015 Klaus Hartl & Fagner Brack
- * Released under the MIT license
- */
+* JavaScript Cookie v2.1.0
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
@ericntd Thanks for the Pull Request!

JavaScript Cookie is (quote from description):

> A simple, lightweight JavaScript API for handling **browser** cookies

One should not expect that it works inside node, but in environments that contain a `document`. Can you please explain your use case better so that I understand what exactly is the problem?

ericntd:
@FagnerMartinsBrack I call `Cookies.get()`, `Cookies.set()` etc. in a lot of places.

So my choices now either:  
I need to update my code everywhere I call your library - check if the environment is browser is Node JS server and only call your library when it's on the browser side
or
Your library should check whether it can get the `document` or not

What do you think?

FagnerMartinsBrack:
@ericntd thanks for the response!

One thing that is still not clear is: why are you calling JavaScript Cookie in the server-side if it is not meant to be used in the server side? Isn't this a problem that should be addressed in your own usage of JavaScript Cookie instead of changing the library by adding additional features for everyone that is already using it?

Maybe a code snippet will help a lot to understand the "why" for this.

I am totally ok in changing the lib scope to be more isomorphic, but we still need at least one real compelling use case that justifies not just the effort, but the bytes that are going to be downloaded by those who already use the library in the browser.

Looking forward on your feedback.

ericntd:
@FagnerMartinsBrack my code is supposed to be isomorphic which means the same piece of code supposed to work on both browser and server side.

When I started out I didn't have that mind set, I was only trying to make it work on browser side.

Somehow I don't have a good feeling about adding an `if else` statement to my code to check the environment

FagnerMartinsBrack:
@ericntd Just confirm then. The library itself does not do anything outside a browser environment, it is just noop. The thing is that you just want that it doesn't throw any errors if it is running in node so that you can use the same codebase in many environments. Is that it?

ericntd:
yes, sir @FagnerMartinsBrack you summed it up perfectly

FagnerMartinsBrack:
It might make sense doing something like `document = {}` inside the cookie scope and see if the increase of bytes is relevant. I don't think it is.

We need to reduce the amount of changes in this PR and add a test for that inside the node environment, testing that the use of any of it's APIs will not cause an exception to be thrown.

Are you willing to do that so we can see if we can pull this through?

src/js.cookie.js
```diff
@@ -57,7 +57,7 @@
 
 				if (!converter.write) {
 					value = encodeURIComponent(String(value))
-						.replace(/%(23|24|26|2B|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g, decodeURIComponent);
```
FagnerMartinsBrack: Unnecessary spacing change.

src/js.cookie.js
```diff
@@ -1,10 +1,10 @@
...
- *
- * Copyright 2006, 2015 Klaus Hartl & Fagner Brack
- * Released under the MIT license
- */
+* JavaScript Cookie v2.1.0
+* https://github.com/js-cookie/js-cookie
+*
+* Copyright 2006, 2015 Klaus Hartl & Fagner Brack
+* Released under the MIT license
+*/
```
FagnerMartinsBrack: Unnecessary spacing change.

src/js.cookie.js
```diff
@@ -99,8 +103,8 @@
 
 				try {
 					cookie = converter.read ?
-						converter.read(cookie, name) : converter(cookie, name) ||
-						cookie.replace(rdecode, decodeURIComponent);
+					converter.read(cookie, name) : converter(cookie, name) ||
+					cookie.replace(rdecode, decodeURIComponent);
```
FagnerMartinsBrack: Unnecessary spacing change.

src/js.cookie.js
```diff
@@ -66,13 +66,17 @@
...
+					return (document.cookie = [
+						key, '=', value,
+						attributes.expires && '; expires=' + attributes.expires.toUTCString(),// use expires attribute, max-age is not supported by IE
+						attributes.path    && '; path=' + attributes.path,
+						attributes.domain  && '; domain=' + attributes.domain,
+						attributes.secure ? '; secure' : ''
+					].join(''));
+				} else {
+					return '';
+				}
```
FagnerMartinsBrack: We can do something like `var document = {}` in this scope, so that it will be noop for read and write operations.

src/js.cookie.js
```diff
@@ -66,13 +66,17 @@
...
 				key = key.replace(/[\(\)]/g, escape);
 
-				return (document.cookie = [
-					key, '=', value,
-					attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
-					attributes.path    && '; path=' + attributes.path,
-					attributes.domain  && '; domain=' + attributes.domain,
-					attributes.secure ? '; secure' : ''
-				].join(''));
+				if (typeof document !== 'undefined') {
```
FagnerMartinsBrack: We can do something like `var document = {}` in this scope, so that it will be noop for read and write operations. In this case there will be no need to use `typeof`and `!==` and `undefined`, which are unminifiable tokens.

ericntd:
Hi @FagnerMartinsBrack just to clarify, you are asking the other js-cookie members right?

FagnerMartinsBrack:
> Hi @FagnerMartinsBrack just to clarify, you are asking the other js-cookie members right?

@ericntd You are the one who opened the Pull Request and will directly benefit from this change. Shouldn't I assume that you are willing to work on this feature?

After all, this is the whole purpose of open-source :smiley:

FagnerMartinsBrack:
I can help you with whatever you want to work on this :+1:

ericntd:
@FagnerMartinsBrack of course, how do I proceed (I've never done this before)?
I should make the changes you pointed out, commit and make another pull request right?

FagnerMartinsBrack:
Take a look in [this article](https://github.com/dchelimsky/rspec/wiki/Topic-Branches).

Ideally you should have created a branch in your fork and then send the Pull Request from that branch (instead of master).

The flow is basically like this:

* `git clone your-fork-url`
* if `git branch` shows you on master, do `git checkout -b name-of-the-branch` to create a topic branch
* Make the changes, `git add .` to add everything to the staging area and `git commit -m "Commit message"`
* Then you push to your fork using `git push origin name-of-your-branch` ("origin" is the name of your "remote" (`git remote` to see the list of all remotes))
* Then go to your fork and create the Pull Request

After creating the Pull Request, any changes to the branch will update the Pull Request too.

Take a look here: https://help.github.com/articles/using-pull-requests/

src/js.cookie.js
ericntd: I've used Atom to auto-indent the code, I'm not particular about spacing here though, will revert

FagnerMartinsBrack:
Creating a new PR using a topic branch instead?

PR #161, opened 2016-03-24
Undefined document patch 1 1

ericntd:


src/js.cookie.js
```diff
@@ -34,6 +34,7 @@
 	function init (converter) {
 		function api (key, value, attributes) {
 			var result;
+			var document = {};
```
FagnerMartinsBrack: First of all I suggest you create a test first following the patterns of the [common.js](https://github.com/js-cookie/js-cookie/blob/d82dc72dab240ac64756afa7182b0d71e112dba1/test/commonjs.js) test, which checks the commonjs support in a node environment. Then you can jumo to the implementation below.

[What I meant](https://github.com/js-cookie/js-cookie/pull/159#issuecomment-200280852) by ...

> It might make sense doing something like var document = {} inside the cookie scope and see if the increase of bytes is relevant. I don't think it is.

... was not just to add the variable inside the scope, that was just an example expressing the idea. To ensure this won't break other environments (such as the browser), you need to verify first that a document does not exist using [feature detection](http://stackoverflow.com/a/20105161/1400037).

**Example**

```javascript
var document = window.document || {};
```

That will make sure the placeholder object is just created when not running in the browser (The browser has a document, node doesn't). Actually, node doesn't even have a `window`, so we need to check for that too.

Ideally we should test other ways to do it to find the most optimal one. The project has a task called `grunt compare-size` (see docs [here](https://github.com/rwaldron/grunt-compare-size)), it compares the gzipped size of the code between branches so that we can find the one that contains the best solution calculating the trade-offs between code size and performance.

See the [Contributing Guidelines](https://github.com/js-cookie/js-cookie/blob/master/CONTRIBUTING.md#getting-started). In case you are having trouble to run the project even after reading the guidelines, please let me know  so that we can improve those docs!

FagnerMartinsBrack:
Closing in favor of #162.

PR #162, opened 2016-03-24
added the check for undefined document

ericntd:
I've already ran `./travis.sh` on local.

The `var document = {}` makes Qunit tests failed so it cannot be used

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -66,13 +66,13 @@
 				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
 				key = key.replace(/[\(\)]/g, escape);
 
-				return (document.cookie = [
+				return typeof document !== 'undefined'? (document.cookie = [
 					key, '=', value,
 					attributes.expires && '; expires=' + attributes.expires.toUTCString(), // use expires attribute, max-age is not supported by IE
 					attributes.path    && '; path=' + attributes.path,
 					attributes.domain  && '; domain=' + attributes.domain,
 					attributes.secure ? '; secure' : ''
-				].join(''));
+				].join('')): '';
 			}
 
 			// Read
@@ -84,7 +84,7 @@
 			// To prevent the for loop in the first place assign an empty array
 			// in case there are no cookies at all. Also prevents odd result when
 			// calling "get()"
-			var cookies = document.cookie ? document.cookie.split('; ') : [];
+			var cookies = typeof document !== 'undefined' && document.cookie ? document.cookie.split('; ') : [];
 			var rdecode = /(%[0-9A-Z]{2})+/g;
 			var i = 0;
 


FagnerMartinsBrack:
@ericntd Thank you very much for your effort! I will take a look on #161 to see why it failed.

FagnerMartinsBrack:
There was no need to create a new Pull Request though. You can comment in the old Pull Request and then [squash](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) the changes so the comments will be collapsed.

To squash the changes you need to:

1. Find the revision that you want to reset to (the one you want to stay in the history). In this case it is the master revision (hash) in which the branch was created from: d82dc72dab240ac64756afa7182b0d71e112dba1
2. Use `git reset --soft d82dc72dab240ac64756afa7182b0d71e112dba1`
3. Use `git status` to verify that all your changes are in the staging area to b commited.
4. Run `git commit -m "new message"` to create a single commit with all the changes together (all commits in one)
5. Run `git push origin my-branch-name:my-branch-name --force` to update the Pull Request (note the `--force`, because we are editing the history and therefore git needs an additional check to make sure you are doing what you mean to)

FagnerMartinsBrack:
I have created a comment [here](https://github.com/js-cookie/js-cookie/pull/161#discussion_r57312297) explaining why it failed. We should probably close one of the Pull Requests and focus in a single one. There is no need to create multiple Pull Requests for the same feature.

FagnerMartinsBrack:
Ping @ericntd

ericntd:
@FagnerMartinsBrack sorry but I have no absolutely no idea how to go about the writing automated test for this library.

In the what you suggested might lead more changes and I'm not sure if I will be able to over the issue `typeof ...` not minifiable anyway.

ericntd:
frankly, I'm considering closing this merge request and
switching to another library
or
wrapping your library

FagnerMartinsBrack:
> frankly, I'm considering closing this merge request and
> switching to another library
> or
> wrapping your library

It feels that you are blocked by this, but you actually don't need to. You can easily workaround this in your code before this is fixed using something as simples as:

```
function setCookie() {
  if ( typeof document !== "undefined" ) {
    Cookies.set.apply(Cookies, arguments);
  }
}
setCookie('name');
```

When this is fixed, you can just upgrade js-cookie and remove the code.

As far as I know, no other browser cookie library allows running the code in an environment other than the browser without throwing an `Error`. I don't see how switching to another library will fix it.

Since you are having trouble to implement this by yourself I will close this and create a new issue to handle it: #164. Thanks for the effort anyway, if you want to continue this just drop a comment that I will reopen it :+1: .

FagnerMartinsBrack:
@ericntd I have create a [Pull Request](https://github.com/js-cookie/js-cookie/pull/165) fixing this. Please, take a look in [this code](https://github.com/js-cookie/js-cookie/blob/noop-node/src/js.cookie.js) and confirm if the problem is fixed.

FagnerMartinsBrack:
Ping @ericntd.

Can you please use [this modified version of the library](https://github.com/js-cookie/js-cookie/blob/noop-node/src/js.cookie.js) in your code and confirm that the problem is fixed?

I am waiting for your response.

ericntd:
@FagnerMartinsBrack I confirm that [the code mentioned](https://github.com/js-cookie/js-cookie/blob/noop-node/src/js.cookie.js) fixed my issue.

Really appreciate your help.

PR #171, opened 2016-04-08
Consider those who doesn't use semicolons

icamys:
Avoiding "intermediate value" error when used with other modules.
See http://stackoverflow.com/questions/23370269/jquery-autosize-plugin-error-intermediate-value-is-not-a-function

### DIFF
--- a/test/missing_semicolon.html
+++ b/test/missing_semicolon.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title></title>
+    <link href="../node_modules/qunitjs/qunit/qunit.css" rel="stylesheet">
+    <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
+    <script src="utils.js"></script>
+    <script>
+        (function() {
+            var contents = window.loadFileSync('../src/js.cookie.js');
+
+            if (contents !== null) {
+                var script = document.createElement('script');
+                script.innerHTML = '(function (){ return {}; })() ' + contents;
+                document.getElementsByTagName('head')[0].appendChild(script);
+            }
+        })();
+    </script>
+</head>
+<body>
+
+</body>
+</html>
--- a/test/tests.js
+++ b/test/tests.js
@@ -63,6 +63,25 @@ QUnit.test('malformed cookie value in IE', function (assert) {
 	document.body.appendChild(iframe);
 });
 
+// github.com/js-cookie/js-cookie/pull/171
+QUnit.test('missing leading semicolon', function (assert) {
+	assert.expect(1);
+	var done = assert.async();
+	// Sandbox in an iframe so that we can poke around with document.cookie.
+	var iframe = document.createElement('iframe');
+	var loadedSuccessfully = true;
+	iframe.src = 'missing_semicolon.html';
+
+	addEvent(iframe, 'load', function () {
+		iframe.contentWindow.onerror = function () {
+			loadedSuccessfully = false;
+		};
+		assert.strictEqual(loadedSuccessfully, true, "can't throw Object is not a function error");
+		done();
+	});
+	document.body.appendChild(iframe);
+});
+
 QUnit.test('Call to read all when there are cookies', function (assert) {
 	Cookies.set('c', 'v');
 	Cookies.set('foo', 'bar');
--- a/test/utils.js
+++ b/test/utils.js
@@ -111,4 +111,12 @@
 			setCookie: setCookie
 		};
 	};
+
+	window.loadFileSync = function (path) {
+		var xhr = new XMLHttpRequest();
+		xhr.open('GET', path, false);
+		xhr.send(null);
+		return xhr.status === 200 ? xhr.responseText : null;
+	};
+
 }());
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -10,6 +10,7 @@
 		"unescape": true,
 		"lifecycle": true,
 		"using": true,
-		"addEvent": true
+		"addEvent": true,
+	  	"loadFileSync": true
 	}
 }


FagnerMartinsBrack:
Thank you, can you please add a test that reproduces the problem when js-cookie is included after another JavaScript file that doesn't use semicolons?

icamys:
Here's an example:
[js-cookie-test.html.zip](https://github.com/js-cookie/js-cookie/files/211414/js-cookie-test.html.zip)

In my project I use [gulp](http://gulpjs.com/) and [asset builder](https://github.com/austinpray/asset-builder). As you probably know, when we execute gulp command, firstly all 'main' files from bower_components directory are concatenated, after that all client code is appended to it.

I found this problem, when during the concatination js-cookie main file was appended after [JavaScript Canvas to Blob](https://github.com/blueimp/JavaScript-Canvas-to-Blob) main file.

FagnerMartinsBrack:
@icamys Sorry, but I meant a test [here](https://github.com/js-cookie/js-cookie/tree/0c8f5e1fe35da6c1105ac943b3a9a69a757c6d0a/test). We need a test that runs when the build runs on CI to prevent regressions. Even though it is just a semicolon, we need a testable example for posterity.

The [Contributing Guidelines](https://github.com/js-cookie/js-cookie/blob/0c8f5e1fe35da6c1105ac943b3a9a69a757c6d0a/CONTRIBUTING.md#pull-requests) shows how to contribute to js-cookie, and there it is said that we need a test when proposing a feature through a Pull Request.

Can you do that and amend on this PR? Otherwise you can create an issue about this.

icamys:
@FagnerMartinsBrack Today I've added a test, but I'm getting a TypeError from PhantomJS:

```
Testing http://127.0.0.1:9998/ .......>> PhantomJS error:
>>  TypeError: Object is not a function (evaluating '(function(){ return {}; })()') 
>>  [ { file: 'http://127.0.0.1:9998/missing_semicolon.html',
>>     line: 1,
>>     function: 'global code' },
>>   { file: '', line: 0, function: 'appendChild' } ]
F...................................
>> read - missing leading semicolon
>> Message: encounters `(intermediate value)...` error
>> Actual: undefined
>> Expected: true
>> http://127.0.0.1:9998/tests.js:75:21
```

I'm virgin with PhantomJS, that's why I don't really know how to fix this error, but it's definitely a real situation, that reproduces the problem.

The test is pretty simple. In HTML using ajax request I'm getting the contents of js-cookie.js file, then prepending to it an expression that reproduces the said error:

```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title></title>
    <script>
        var xmlHttp = new XMLHttpRequest();
        xmlHttp.open( "GET", '../src/js.cookie.js', false );
        xmlHttp.send( null );
        window.scriptContent = '(function (){ return {}; })() ';
        window.scriptContent += xmlHttp.responseText;
        window.scriptContent += ';window.ok = true;';

        var script = document.createElement('script');
        script.innerHTML = window.scriptContent;

        document.getElementsByTagName('head')[0].appendChild(script);
    </script>
</head>
<body>
</body>
</html>
```

If you can give me the cue how to bypass this error, I would be grateful.

By the way, here is the QUnit test itself:

```
// github.com/js-cookie/js-cookie/pull/171
QUnit.test('missing leading semicolon', function (assert) {
	assert.expect(1);
	var done = assert.async();
	// Sandbox in an iframe so that we can poke around with document.cookie.
	var iframe = document.createElement('iframe');
	iframe.src = 'missing_semicolon.html';

	addEvent(iframe, 'load', function () {
		assert.strictEqual(iframe.contentWindow.ok, true, 'encounters `(intermediate value)...` error');
		done();
	});
	document.body.appendChild(iframe);
});
```

FagnerMartinsBrack:
@iamys There is no need to, you can use that test in this Pull Request, maybe just changing the message to explain what is going on. js-cookie is trying to execute the Object as a function because of the lack of semicolon, if you add the semicolon in the beginning of the file, it will pass.

Can you `git amend` this Pull Request with those tests so that I can review properly using Github diff?

Thank you!

icamys:
@FagnerMartinsBrack I've amended this PR.

> js-cookie is trying to execute the Object as a function because of the lack of semicolon, if you add the semicolon in the beginning of the file, it will pass.

Agree, though we can't be sure about the code, that is executed before the module from time to time. That's why there is a need for an extra semicolon.

FagnerMartinsBrack:
> though we can't be sure about the code, that is executed before the module from time to time

I don't understand what you mean, could you please elaborate? Thanks.

test/missing_semicolon.html
```diff
@@ -0,0 +1,23 @@
...
+<head>
+    <meta charset="UTF-8">
+    <title></title>
+    <script>
+        var xmlHttp = new XMLHttpRequest();
+        xmlHttp.open( "GET", '../src/js.cookie.js', false );
+        xmlHttp.send( null );
+        window.scriptContent = '(function (){ return {}; })() ';
+        window.scriptContent += xmlHttp.responseText;
+        window.scriptContent += ';window.ok = true;';
```
FagnerMartinsBrack: Is there really a reason for changing `window.ok = true` after it is loaded? Can't we just check if the test throws an `Error`?

test/missing_semicolon.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title></title>
+    <script>
+        var xmlHttp = new XMLHttpRequest();
```
FagnerMartinsBrack: Can we abstract the AJAX request into a `loadFile(fileName)` function that accepts a NodeFunction (a callback that returns an error in the first argument and the content in the second)?

```javascript
loadFile('../src/js.cookie.js', function loadedFile(error, libraryContent) {
  if (error) { throw error; }
  // rest of the code here
});
```

test/missing_semicolon.html
```diff
@@ -0,0 +1,23 @@
...
+        xmlHttp.open( "GET", '../src/js.cookie.js', false );
+        xmlHttp.send( null );
+        window.scriptContent = '(function (){ return {}; })() ';
```
FagnerMartinsBrack: We need to create a closure to prevent leaking global variables.

test/missing_semicolon.html
```diff
@@ -0,0 +1,23 @@
...
+        window.scriptContent += xmlHttp.responseText;
+        window.scriptContent += ';window.ok = true;';
+
+        var script = document.createElement('script');
```
FagnerMartinsBrack: Same as [this comment](https://github.com/js-cookie/js-cookie/pull/171/files#r59530280)

test/missing_semicolon.html
FagnerMartinsBrack: Same as [this comment](https://github.com/js-cookie/js-cookie/pull/171/files#r59530280)

FagnerMartinsBrack: The function can go in the [test utils file](https://github.com/js-cookie/js-cookie/blob/4a4122eb06bc6c4fa8034f00dd9170bd07d4ce7c/test/utils.js).

test/tests.js
```diff
@@ -63,6 +63,21 @@ QUnit.test('malformed cookie value in IE', function (assert) {
...
+// github.com/js-cookie/js-cookie/pull/171
+QUnit.test('missing leading semicolon', function (assert) {
+	assert.expect(1);
+	var done = assert.async();
+	// Sandbox in an iframe so that we can poke around with document.cookie.
+	var iframe = document.createElement('iframe');
+	iframe.src = 'missing_semicolon.html';
+
+	addEvent(iframe, 'load', function () {
+		assert.strictEqual(iframe.contentWindow.ok, true, 'encounters `(intermediate value)...` error');
```
FagnerMartinsBrack: Since we are not simulating the `(intermediate value)` error specifically, what about creating a more specific message like "can't throw Object is not a function error"?

test/missing_semicolon.html
FagnerMartinsBrack: Oh yeah you can't, since it is inside an iframe xD, correct me if I am wrong.

FagnerMartinsBrack: In this case I guess something like `window.jsCookieLoadedSuccessfully` will explain what this `ok` means when reading the test. Not a big deal though.

FagnerMartinsBrack:
@icamys Thank you very much for the quality of this Pull Request :O. I have made some comments, looking forward on your feedback.

test/missing_semicolon.html
FagnerMartinsBrack: I found a good resource in how to get errors from iframe in case you are interested: http://stackoverflow.com/a/14041297

icamys:
@FagnerMartinsBrack I've updated PR again considering your comments.

test/tests.js
```diff
@@ -63,6 +63,25 @@ QUnit.test('malformed cookie value in IE', function (assert) {
...
+	// Sandbox in an iframe so that we can poke around with document.cookie.
+	var iframe = document.createElement('iframe');
+	var loadedSuccessfully = true;
+	iframe.src = 'missing_semicolon.html';
+
+	addEvent(iframe, 'load', function () {
+		iframe.contentWindow.onerror = function () {
+			loadedSuccessfully = false;
+		};
+		assert.strictEqual(loadedSuccessfully, true, 'can\'t throw Object is not a function error');
```
FagnerMartinsBrack: We usually don't escape single quotes, we use double quotes instead (yeah, pretty crazy, but the code was like this since forever xD)

test/utils.js
```diff
@@ -111,4 +111,27 @@
...
 	};
+
+	/**
+	 * Loads file contents via ajax GET request and passes it to callback function
+	 * @param {string} path Path to file
+	 * @param {function} callback Function that accepts two arguments:
+	 *                            first one - ajax error message,
+	 *                            second one - file contents.
+	 * @returns {boolean}
+	 */
```
FagnerMartinsBrack: Since this code is sync can we use `loadFileSync` name and return the contents instead of using a callback? Also, these comments are unnecessary, the behavior van be inferred through the usage. Also, the boolean return is not being used, no need to implement that.

icamys:
@FagnerMartinsBrack I still hope that PR would be accepted =)

FagnerMartinsBrack:
Ugh! It seems I made a mistake with the double quotes! see [here](https://github.com/js-cookie/js-cookie/blob/4a4122eb06bc6c4fa8034f00dd9170bd07d4ce7c/test/tests.js#L240), we are not using double quotes anymore it seems =(.

Can you please revert that, to fix the code style error on CI? Sorry for that =/
Yeah, this PR will be accepted for v2.1.1, no worries :)

FagnerMartinsBrack:
Nevermind, I can fix this. Thank you!

Issue #177, opened 2016-04-13
Using js-cookie without commonjs or amd


MartinMuzatko:
Hello there! Is there currently a way to retrieve js-cookie on the clientside without any commonjs or amd?
If so, please document.
If not, please document how to install js-cookie.

MartinMuzatko:
Figured it out! Clearer info would be very much appreciated never the less

FagnerMartinsBrack:
> Clearer info would be very much appreciated never the less

@MartinMuzatko We are aways looking for feedback in the docs. What is the problem with how it is documented in the README?

<img width="1047" alt="screenshot 2016-04-13 22 25 58" src="https://cloud.githubusercontent.com/assets/835857/14493080/d6b11918-01c6-11e6-8d87-aa579b815f2e.png">

Looking forward on your feedback.

MartinMuzatko:
Oh I think I overlooked that. Maybe I'm not used to getting a working answer so fast. Haha. So by no means is it your fault.

I think only these big words confused me a bit.

> js-cookie supports npm and Bower under the name js-cookie
>It can also be loaded as an AMD or CommonJS module.

I thought you would require me to use AMD or CommonJS together with npm/bower.

Alternatively, you could destroy their implied context by saying:

"js-cookie can be found as `js-cookie` on npm and Bower."

And maybe move it up over the rawgit attention note.
Additionaly, you could improve the way to get this software with:

"js-cookie can be found as `js-cookie` on npm and Bower or downloaded here alternatively"

FagnerMartinsBrack:
What about:

> js-cookie can also be found on npm and Bower under the name `js-cookie`, or downloaded here alternatively.

Or better, we could separate in sub-sections inside "Installation":

Installation
  - Browser
  - Package Managers
  - Module Loaders

Which one you think would be better?

MartinMuzatko:
I like both ideas. However the last one, separating the ways to install it into their use cases is a great way to offer information. 
I'm not sure about the section "Browser". Isn't it only used with the browser in the end? I know with Browser you probably mean "download it manually and use a script tag". I think there could be a better title.

Examples:
https://jquery.com/download/
http://riotjs.com/download/

"Direct Download" is a good way to describe it.

FagnerMartinsBrack:
[...]

PR #182, opened 2016-04-22
Separate the installation docs in sections

FagnerMartinsBrack:
Problem reported on #177
To view the updated README file, [click here](https://github.com/js-cookie/js-cookie/tree/readme-installation#readme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ A simple, lightweight JavaScript API for handling cookies
 
 ## Installation
 
-Include the script (unless you are packaging scripts somehow else):
+### Direct download
+
+Download the script [here](https://github.com/js-cookie/js-cookie/blob/v2.1.1/src/js.cookie.js) and include it (unless you are packaging scripts somehow else):
 
 ```html
 <script src="/path/to/js.cookie.js"></script>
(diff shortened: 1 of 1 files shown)


MartinMuzatko:
Please add the direct link to to the raw file: 
https://raw.githubusercontent.com/js-cookie/js-cookie/master/src/js.cookie.js

 > **Do not include the script directly from GitHub** (http://raw.github.com/...). The file is being served as text/plain and as such being blocked in Internet Explorer on Windows 7 for instance (because of the wrong MIME type). Bottom line: GitHub is not a CDN.


The problem with the paragraph above is, that users tend to ignore the text coming after "DO NOT" or just skip it, as it is not relevant. Putting the download file within a "do not" paragraph is not very intuitive, which is why I'd word it otherwise.

**Direct Download**

Download the file here: http://raw.github.com/...

**Do not include the script directly from GitHub** (http://raw.github.com/...). The file is being served as text/plain and as such being blocked in Internet Explorer on Windows 7 for instance (because of the wrong MIME type). Bottom line: GitHub is not a CDN.

FagnerMartinsBrack:
@MartinMuzatko Thanks for the feedback! I have made some changes, can you take a look please?

FagnerMartinsBrack:
But the way, looking at it again it doesn't seem worth pointing people out to the master branch, which is generally unstable. Ideally we should be pointing to the latest tagged release instead.

@MartinMuzatko what do you think?

MartinMuzatko:
Good idea! Please update that.

FagnerMartinsBrack:
> Good idea! Please update that.

Which one is a good idea, the relative url or a pointer always to the latest release?

MartinMuzatko:
The latest tag. Thank you for asking again. I could be more precise.

PR #200, opened 2016-05-13
Update the UTF-8 percent encoded regex to match only valid encodings …

mmilford:
…and to ignore case

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-cookie",
-  "version": "2.1.1",
+  "version": "2.1.2",
   "description": "A simple, lightweight JavaScript API for handling cookies",
   "main": "src/js.cookie.js",
   "directories": {
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -88,7 +88,14 @@
 			// in case there are no cookies at all. Also prevents odd result when
 			// calling "get()"
 			var cookies = document.cookie ? document.cookie.split('; ') : [];
-			var rdecode = /(%[0-9A-Z]{2})+/g;
+
+			// This regex matches all valid values for UTF-8 percent encoding.  See https://tools.ietf.org/html/rfc3629
+			// Values U+0000 - U+007F are 1 byte encoded
+			// Values U+0080 - U+07FF are 2 byte encoded.  The first byte begins with 110 ([CD]).  The continuation byte begins with 10 ([89AB])
+			// Values U+8000 - U+FFFF are 3 byte encoded.  The first byte begins with 1110 (E).  The continuation byte begins with 10 ([89AB])
+			// Values U+10000 - U+FFFFF are 4 byte encoded.  The first byte begins with 111100 (F[0-3]).  The continuation byte begins with 10 ([89AB])
+			// Values U+100000 - U+10FFFF are 4 byte encoded.  The first byte is 11110100 (F4).  The second byte begins with 1000 (8).  The continuation byte begins with 10 ([89AB])
+			var rdecode = /(%[0-7][0-9A-F])|(%[CD][0-9A-F]%[89AB][0-9A-F])|(%E[0-9A-F]%[89AB][0-9A-F]%[89AB][0-9A-F])|(%F3%[89AB][0-9A-F]%[89AB][0-9A-F]%[89AB][0-9A-F])|(%F4%8[0-9A-F]%[89AB][0-9A-F]%[89AB][0-9A-F])+/gi;
 			var i = 0;
 
 			for (; i < cookies.length; i++) {
--- a/test/tests.js
+++ b/test/tests.js
@@ -41,6 +41,12 @@ QUnit.test('percent character in cookie value mixed with encoded values', functi
 	assert.strictEqual(Cookies.get('bad'), 'foo%bar"baz%bax=', 'should read the percent character');
 });
 
+QUnit.test('percent character in cookie value mixed with encoded values lower case', function (assert) {
+	assert.expect(1);
+	document.cookie = 'mybad=foo%bar%22baz%bax%3d';
+	assert.strictEqual(Cookies.get('mybad'), 'foo%bar"baz%bax=', 'should read the percent character');
+});
+
 // github.com/carhartl/jquery-cookie/pull/88
 // github.com/carhartl/jquery-cookie/pull/117
 QUnit.test('malformed cookie value in IE', function (assert) {
@@ -96,16 +102,6 @@ QUnit.test('RFC 6265 - reading cookie-octet enclosed in DQUOTE', function (asser
 	assert.expect(1);
 	document.cookie = 'c="v"';
 	assert.strictEqual(Cookies.get('c'), 'v', 'should simply ignore quoted strings');
-});
-
-// github.com/js-cookie/js-cookie/pull/62
-QUnit.test('Call to read cookie when there is another unrelated cookie with malformed encoding in the value', function (assert) {
-	assert.expect(2);
-	document.cookie = 'invalid=%A1';
-	document.cookie = 'c=v';
-	assert.strictEqual(Cookies.get('c'), 'v', 'should not throw a URI malformed exception when retrieving a single cookie');
-	assert.deepEqual(Cookies.get(), { c: 'v' }, 'should not throw a URI malformed exception when retrieving all cookies');
-	Cookies.withConverter(unescape).remove('invalid');
 });
 
 // github.com/js-cookie/js-cookie/issues/145


test/tests.js
```diff
@@ -98,16 +104,6 @@ QUnit.test('RFC 6265 - reading cookie-octet enclosed in DQUOTE', function (asser
...
-// github.com/js-cookie/js-cookie/pull/62
-QUnit.test('Call to read cookie when there is another unrelated cookie with malformed encoding in the value', function (assert) {
-	assert.expect(2);
-	document.cookie = 'invalid=%A1';
-	document.cookie = 'c=v';
-	assert.strictEqual(Cookies.get('c'), 'v', 'should not throw a URI malformed exception when retrieving a single cookie');
-	assert.deepEqual(Cookies.get(), { c: 'v' }, 'should not throw a URI malformed exception when retrieving all cookies');
-	Cookies.withConverter(unescape).remove('invalid');
-});
-
```
FagnerMartinsBrack: Why removing this test?

test/tests.js
```diff
@@ -41,6 +41,12 @@ QUnit.test('percent character in cookie value mixed with encoded values', functi
 	assert.strictEqual(Cookies.get('bad'), 'foo%bar"baz%bax=', 'should read the percent character');
 });
 
+QUnit.test('percent character in cookie value mixed with encoded values lower case', function (assert) {
+	assert.expect(1);
+	document.cookie = 'mybad=foo%bar%22baz%bax%3d';
```
FagnerMartinsBrack: In `3d`, see [here](https://github.com/js-cookie/js-cookie/commit/6b66ed4e29dbfa1a223804cbdb8169e3ea636b52#commitcomment-12300704) for some context:

Although lowercase and uppercase digits are equivalent per spec, is there any browser where those characters are treated differently after being encoded with `encodeURIComponent`? Or what is the use case for considering lowercase characters in percent-encoding for the default decoding mechanism, can't this be done sing [converters](https://github.com/js-cookie/js-cookie#converters)?

```javascript
Cookies = Cookies.withConverter(acceptLowercaseEncoding);
```

We use the [Strictness Principle](https://medium.com/@fagnerbrack/the-strictness-principle-9997e483cafb) for new features in the library, therefore we need at least a reasonable use case that justifies increasing the scope of the behavior of the default decoding mechanism.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "js-cookie",
-  "version": "2.1.1",
+  "version": "2.1.2",
```
FagnerMartinsBrack: There is no need to increase the version in this commit, this is done when a new version is released.

src/js.cookie.js
```diff
@@ -88,7 +88,14 @@
...
 			var cookies = document.cookie ? document.cookie.split('; ') : [];
-			var rdecode = /(%[0-9A-Z]{2})+/g;
+
+			// This regex matches all valid values for UTF-8 percent encoding.  See https://tools.ietf.org/html/rfc3629
+			// Values U+0000 - U+007F are 1 byte encoded
+			// Values U+0080 - U+07FF are 2 byte encoded.  The first byte begins with 110 ([CD]).  The continuation byte begins with 10 ([89AB])
+			// Values U+8000 - U+FFFF are 3 byte encoded.  The first byte begins with 1110 (E).  The continuation byte begins with 10 ([89AB])
+			// Values U+10000 - U+FFFFF are 4 byte encoded.  The first byte begins with 111100 (F[0-3]).  The continuation byte begins with 10 ([89AB])
+			// Values U+100000 - U+10FFFF are 4 byte encoded.  The first byte is 11110100 (F4).  The second byte begins with 1000 (8).  The continuation byte begins with 10 ([89AB])
+			var rdecode = /(%[0-7][0-9A-F])|(%[CD][0-9A-F]%[89AB][0-9A-F])|(%E[0-9A-F]%[89AB][0-9A-F]%[89AB][0-9A-F])|(%F3%[89AB][0-9A-F]%[89AB][0-9A-F]%[89AB][0-9A-F])|(%F4%8[0-9A-F]%[89AB][0-9A-F]%[89AB][0-9A-F])+/gi;
```
FagnerMartinsBrack: The default decoding mechanism for cookie values is meant to be used from js-cookie to js-cookie. Anything different should be done using [converters](https://github.com/js-cookie/js-cookie#converters).

Can you please create some tests in the [encoding.js](https://github.com/js-cookie/js-cookie/blob/f2b25713e8fae658d418b60710adc0e0d14998d9/test/encoding.js) file that reproduces the problem this code is trying to solve using js-cookie APIs to set the cookie?

If this problem is not about js-cookie, but an attempt to adapt itself to external cookies that are created in another format, can you please explain where are you getting this cookie so that we can understand what the use case is? Or if possible, create a test simulating the use case.

package.json
mmilford: Sorry, didn't realize.

test/tests.js
mmilford: No, I'm not aware of any browsers with differing behavior for encodeURIComponent.  However, it's not clear from your documentation that this library can only be used to read cookies that it creates.

FagnerMartinsBrack: > ... it wasn't clear to me from my reading of your documentation that this library can only be used to read cookies that it creates.

Do you want to create a Pull Request that improves the documentation in that sense?

src/js.cookie.js
mmilford: The cookie's being created by an ASP.NET MVC web project, using a standard method (UrlEncode) https://msdn.microsoft.com/en-us/library/system.web.httputility.urlencode(v=vs.110).aspx.  Why they chose to lowercase the hex values is unclear, but as you noted above it is spec compliant.

As I mentioned before, it wasn't clear to me from the documentation that I'd have to write custom converters for use cases that weren't strictly js-cookie to js-cookie.  However, as a consumer of this library I think that it would make sense for it to be updated to read all spec compliant cookies, as that seems like a standard use case.  Custom converters seems appropriate for custom cases, which I would argue this is not.

I can definitely add some unit tests in encoding.js to highlight what are valid and invalid UTF-8 percent encodings.

FagnerMartinsBrack: > The cookie's being created by an ASP.NET MVC web project, using a standard method (UrlEncode) https://msdn.microsoft.com/en-us/library/system.web.httputility.urlencode(v=vs.110).aspx. Why they chose to lowercase the hex values is unclear, but as you noted above it is spec compliant.

I understand. The problem is that the amount of unminifiable content that is added in the code will make it bigger than 900 bytes gzipped, so I am definitely not willing to add such amount of code if the problem can be fixed using converters. Out of the box, js-cookie supports only what it produces, maybe we can update the docs to make that clear.

I don't use ASP.NET, therefore, can you add the peculiarities of ASP.NET in the [SERVER_SIDE](https://github.com/js-cookie/js-cookie/blob/f2b25713e8fae658d418b60710adc0e0d14998d9/SERVER_SIDE.md) documentation along with the proper converter to make it handle ASP.NET cookie, just like what was done with the PHP one? That would be much appreciated!

FagnerMartinsBrack: As a sidenote, in order to provide what js-cookie is meant to provide (an API for handling cookie), I would argue that we didn't even needed a standard decoding/encoding mechanism, each use case could use converters. The problem is that many people use only js-cookie and anything else, that is why making it just work by default is a must.

test/tests.js
FagnerMartinsBrack: @mmilford Do you think we can improve the docs about this? Do you have any suggestion? Thanks!

src/js.cookie.js
FagnerMartinsBrack: @mmilford Can you provide an example with code that make the cookies work for ASP.NET using converters? I can add that to the [SERVER_SIDE](https://github.com/js-cookie/js-cookie/blob/f2b25713e8fae658d418b60710adc0e0d14998d9/SERVER_SIDE.md) integration file.

Thanks!

FagnerMartinsBrack:
Ping @mmilford.

test/tests.js
FagnerMartinsBrack: I have created an issue for this, set milestone 2.1.3: https://github.com/js-cookie/js-cookie/issues/211

src/js.cookie.js
FagnerMartinsBrack: If you want to help us improving the docs I recommend creating a Pull Request. I will be closing this for now because it doesn't seems to have been tracked by the [OP](https://medium.com/@fagnerbrack/what-does-op-means-on-github-8db958b604af) anymore.

FagnerMartinsBrack:
Closing due to the lack of feedback from the [OP](https://medium.com/@fagnerbrack/what-does-op-means-on-github-8db958b604af). If you want to continue the discussion, please comment here mentioning one of the members.

SHUAIB1244:
|
  | <input type="hidden" name="merchant_identifier" value="abKSBKKe" id="id_merchant_identifier">
  | <input type="hidden" name="merchant_reference" value="Appointment-11603202349896711-vXLCYMgdP6" id="id_merchant_reference">
  | <input type="hidden" name="language" value="en" id="id_language">
  | <input type="hidden" name="signature" value="5d3a6b7f8353e6665091aae82779255d3d5b8c36c11f49325083b90a8d519a7819d5b9e7657f18f57bdf1b9cd021bb68a8230bb513c418bbff216b1a00658d6c" id="id_signature">

PR #204, opened 2016-05-23
Correct notice on attributes in remove function

bendiksolheim:
The notice stated that path, domain and secure attributes needs to be the same
as when the cookie was created when removing a cookie, but in reality only the
path and domain attributes needs to be the same. This commit removes "secure"
from this notice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Cookies.remove('name'); // fail!
 Cookies.remove('name', { path: '' }); // removed!
 ```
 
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
That is a [freaking awesome commit](https://medium.com/@fagnerbrack/one-commit-one-change-3d10b10cebbf)!

FagnerMartinsBrack:
Oh, now I realized, it is not an awesome commit because you changed `were` -> `was`, not that it really matters in practice anyway ;)

bendiksolheim:
Ahh, so close! I agree with you though, smaller commit without any noise is definitely preferable :) I'll try to suppress my inner grammar voice the next time ;)

Issue #206, opened 2016-06-03
Broken behavior after uglifying code with unsafe=true


everdimension:
This might be an edge case, but I want to let this issue be known.

This library's behavior breaks after passing its code through and [uglifier](https://github.com/mishoo/UglifyJS2) plugin with an [`unsafe`](https://github.com/mishoo/UglifyJS2#compressor-options) option set to `true`.

This is what the uglifier does to the lines [72-78](https://github.com/js-cookie/js-cookie/blob/v2.1.2/src/js.cookie.js#L72-L78):
![screen shot 2016-06-03 at 16 10 16](https://cloud.githubusercontent.com/assets/5347023/15779769/fffc0f8c-29a6-11e6-8c2c-9c7b37792c8c.png)

This results in this broken behavior: 

```javascript
jsCookie.set('rnd', '43', { path: '/' });
// returns "rnd=43; path=/undefined"
```

You may argue that that's exactly what should be expected from an "unsafe" option and that this plugin is not responsible for the actions of other libs.

But still, the usage of `uglify` is very common even with the `unsafe` option on.

I'd like to offer a PR in a couple of hours which would fix the issue. You may reject it if you believe that it is not needed, but I think that making this library more robust will only be a good thing and will not harm its behavior in any way.

everdimension:
Open PR https://github.com/js-cookie/js-cookie/pull/207 for this issue.

FagnerMartinsBrack:
@everdimension Thanks for opening the issue.

The code was done like that originally to spare bytes when using uglify in our minification system, so that we still advertise correctly (~800 bytes gzipped). See PR https://github.com/js-cookie/js-cookie/pull/14 and related links for context.

To make sure that this is really necessary, can you please post the following information in the Pull Request?

1. The GZipped size of the file using uglify without the `unsafe` option.
2. The GZipped size of the file using uglify with the `unsafe` option and the code on your Pull Request.

If (1) have more **GZipped** bytes than (2), then we will recommend not using the `unsafe` because it will make no difference. If (1) have less **GZipped** bytes than (2), then we will consider doing something about it, maybe even run the tests on that minified version to make sure it doesn't break in the  future.

If you don't know how to see the GZipped size using our build tool and [grunt-compare-size](https://github.com/rwaldron/grunt-compare-size), please et me know that I will setup a small guide of how to run it in this project to make contributions easier.

everdimension:
@FagnerMartinsBrack 
[...]

PR #207, opened 2016-06-03
Fix concatenation logic: Add support for uglifyJS 'unsafe' option

everdimension:
Fixes https://github.com/js-cookie/js-cookie/issues/206

### DIFF
--- a/test/missing_semicolon.html
+++ b/test/missing_semicolon.html
@@ -8,7 +8,7 @@
     <script src="utils.js"></script>
     <script>
         (function() {
-            var contents = window.loadFileSync('../src/js.cookie.js');
+            var contents = window.loadFileSync('../build/js.cookie.min.js');
 
             if (contents !== null) {
                 var script = document.createElement('script');
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -70,11 +70,15 @@ module.exports = function (grunt) {
 		},
 		uglify: {
 			options: {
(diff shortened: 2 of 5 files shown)


everdimension:
Used gulp-compare-size to get this info:

   | `unsafe` option | raw size | gZipped size
----|--------------------- | ----------- | ----------------
current build | false | 1710 | 894 
current build <sup>bugged behavior</sup> | true | 1697 | 891
build from PR | false | 1716 | 895
build from PR | true | 1703 | 891

So i'd say this PR practically doesn't affect the bundle size but makes it more robust.

FagnerMartinsBrack:
We need to run the tests in the minified file using the `unsafe` option in js-cookie build. Can you do that? If all tests pass and it spares 4 bytes as the table you posted, the this is definitely good to go!

everdimension:
@FagnerMartinsBrack updated the PR with relevatnt changes to the tests

test/missing_semicolon.html
```diff
@@ -21,4 +21,4 @@
 <body>
 
 </body>
-</html>
\ No newline at end of file
+</html>
```
FagnerMartinsBrack: Can we please remove this unrelated change so that the changes are logged only in the relevant lines? Thanks!

everdimension: Right, sorry. I'm using atom which does this by default. I'll undo this and squash the commit.

everdimension: Done ;)

PR #216, opened 2016-06-19
Update QUnit to version 2.0.0

FagnerMartinsBrack:
For some reason the build on IE started failing in Sauce Labs after https://github.com/js-cookie/js-cookie/commit/337419f7366264c2c9bc34ec6f69d9fd434ceb35, need to investigate.

See https://travis-ci.org/js-cookie/js-cookie/builds/138708313#L1035-L1056

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "grunt-jscs": "3.0.0",
     "grunt-saucelabs": "8.6.2",
     "gzip-js": "0.3.2",
-    "qunitjs": "1.23.1",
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
QUnit 2.0.0 removed support for IE 6, 7 and 8. They just [forgot to update the docs](https://github.com/jquery/qunit/issues/1012#issuecomment-228563812).

Since the effort for our support is still trivial for ie 6, 7 and 8, we will be using QUnit v1 for some time.

Issue #233, opened 2016-07-12
Can't import npm package?


hakunin:
Hi,
sorry for a stupid question, but I just can't seem to import the `Cookies` object in my project no matter what I try.

Here is what I tried:
```
import 'js-cookie'
````
and
```
import Cookie from 'js-cookie'
window.Cookie = Cookie
```

None produce any Cookie names object in the window, or local context.
I can import any other packages just fine, but somehow not this one.

Any ideas why that might be?

**Comment by FagnerMartinsBrack** [comment]
We haven't tested the library with any ES6 feature yet. Feel free to create a Pull Request adding compatibility with ES6 while maintaining backwards compatibility with ES5, if that's possible.

Unfortunately, we can't use transpilers in the code on master to use ES6 because many people use the code straight from Github and it might break if their browser does not support ES6 features.

Maybe if you come up with a PR we can talk about it and maybe find the best solution for addressing this.

**Comment by hakunin** [comment]
Oh, in that case it would be nice to have a big warning somewhere so that people don't waste time banging their haid on the desk like I did :)

**Comment by hakunin** [comment]
So what I end up doing was adding a file in my projects where it says
```
window.Cookies = require('js-cookie');
```
And importing it from the main file where I import stuff.

**Comment by FagnerMartinsBrack** [comment]
@hakunin 

> Oh, in that case it would be nice to have a big warning somewhere so that people don't waste time banging their haid on the desk like I did :)

Or maybe it would be nice if someone could help out investigating and implementing the feature, after all this is an open-source project.

The minimum one could do is contributing to make it better as a reward for many hours of confusion and pain that the project saved from, those hours would otherwise be wasted. If someone is wasting time struggling for a feature that doesn't exist, then it might aswell help implementing that :)

**Comment by ThaNarie** [comment]
The es6 module spec doesn't have anything similar to the `module.exports = obj`, there's only default and named exports.

Fortunately, you can still fake it by doing:
```
import * as Cookies from "js-cookie";
```
In this case, all the variable and functions from the exported object will be 'transferred' over the the `Cookies` object.

Note: If you are using TypeScript, use the `--allowSyntheticDefaultImports` flag.

So no code changes needed, adding this usage to the docs however would be nice :)

FagnerMartinsBrack:
@ThaNarie Thanks for the example

jrgleason:
I added 

`import * as Cookies from "js-cookie";`

[...]

PR #237, opened 2016-07-14
fixes #236

### LINKED ISSUE #236: Cookies.get() always returns empty when used with jsdom.
In this block of code:

```js
if (this.json) {
  try {
    cookie = JSON.parse(cookie);
  } catch (e) {}
}
```

If `this` is undefined, it will throw an error causing a `Cookies.get()` to always return an empty `{}`.

The solution is to change the conditional to the following:

```js
if (this && this.json) {
  try {
    cookie = JSON.parse(cookie);
  } catch (e) {}
}
```

This is only failing in my test suite where jsdom is used instead of a browser. Everything seems to work in the browser.

Working on a PR.

lynndylanhurley:
See #236.


FagnerMartinsBrack:
Can we create a test that fails first before fixing the issue to that this use case os covered? Thanks.

lynndylanhurley:
I'm not sure how to go about that without including jsdom as a dev dependency. Are you ok with that?

lynndylanhurley:
Also this is only an issue when using es6 imports. It will take a while to set that up for this package, and it will be a significant change. Is it all really necessary for this patch?

FagnerMartinsBrack:
>I'm not sure how to go about that without including jsdom as a dev dependency. Are you ok with that?

Definitely! Any dependency that helps testing is welcome, we just try to be as strict as possible and just put what is necessary.

> Also this is only an issue when using es6 imports. It will take a while to set that up for this package, and it will be a significant change. Is it all really necessary for this patch?

Can we do that just for this test?

By the way, can we use ES6 imports? See https://github.com/js-cookie/js-cookie/issues/233, we had an issue when someone tried to use he and didn't managed to. Can you provide an example of how to use the library using ES6 modules so that we can maybe document it in the README?

Thanks!

lynndylanhurley:
From that issue:

> Unfortunately, we can't use transpilers in the code on master to use ES6 because many people use the code straight from Github and it might break if their browser does not support ES6 features.

How can we test ES6 imports on node without using transpilers?

FagnerMartinsBrack:
> How can we test ES6 imports on node without using transpilers?

Not sure, how are you using ES6 imports with the current code that is written in ES5?

FagnerMartinsBrack:
Maybe if I understand what is the use case then it will be easier to help, see https://github.com/js-cookie/js-cookie/issues/236#issuecomment-232825987.

lynndylanhurley:
When this code runs in the browser, `this` refers to `window`. That's not the case in node.

jsdom adds a mock version of `window` and everything that it should contain to `global.window` (and `global.document`, etc.). But we can't redefine the global `this`.

We could also change the code to the following, which makes it clear what `this` should refer to in the context that it's being used.

```js
if (window.json) {
  try {
    cookie = JSON.parse(cookie);
  } catch (e) {}
}
```

Would this change be acceptable without including tests? Otherwise we will need to set up babel, mocha, jsdom, and integrate it all into a new grunt task.

FagnerMartinsBrack:
> Also, why is the json property even set to the global window object in the first place? Is that intentional?

That's not intentional and might be a fix that should be done by itself.

> Otherwise we will need to set up babel, mocha, jsdom, and integrate it all into a new grunt task.

Why would we need mocha? We are using QUnit here and we [already have a test for node environment](https://github.com/js-cookie/js-cookie/blob/960a75ed4f5876327c2a7628cc6e41be282effcc/test/node.js).

>  integrate it all into a new grunt task

Can't we just do that only for the tests? Why using Grunt?

As I said, we really need an example that shows the problem so that we can understand how it is being used. Maybe there is a better way to test this, but first I need to understand exactly what is the use case,

package.json
```diff
@@ -36,8 +39,12 @@
     "grunt-contrib-uglify": "1.0.1",
     "grunt-contrib-watch": "1.0.0",
     "grunt-jscs": "3.0.1",
+    "grunt-mocha-test": "^0.12.7",
     "grunt-saucelabs": "8.6.2",
     "gzip-js": "0.3.2",
+    "jsdom": "^9.4.1",
+    "jsdomify": "^2.1.0",
+    "mocha": "^2.5.3",
```
FagnerMartinsBrack: We don't use dynamic ranges.

test/es6-import.js
```diff
@@ -0,0 +1,12 @@
+import jsdomify from 'jsdomify';
+import {assert} from 'chai';
+import Cookies from '../src/js.cookie';
+
+it('should function properly when jsdom is used', () => {
```
FagnerMartinsBrack: Can we reproduce this using QUnit?

Gruntfile.js
```diff
@@ -39,13 +39,21 @@ module.exports = function (grunt) {
 		nodeunit: {
 			all: 'test/node.js'
 		},
+		mochaTest: {
```
FagnerMartinsBrack: Can we run Babel outside our build system, like inside the tests without having to change the Grunt tasks? See the comment: https://github.com/js-cookie/js-cookie/pull/237#issuecomment-232860139.

lynndylanhurley: Babel only runs when mocha is used.

test/es6-import.js
lynndylanhurley: From what I understand, QUnit only runs in the browser. The problem doesn't happen in the browser, only when the code is run serverside in node.

Gruntfile.js
FagnerMartinsBrack: That's weird, isn't Babel just a transpiler and Mocha just the test runner? Why would one have a dependency on the other? Can't you just run Babel standalone?

package.json
lynndylanhurley: Fixed

test/es6-import.js
FagnerMartinsBrack: QUnit runs in the browser and node. If we want to run in node we can use nodeunit. The [example I posted in this comment](https://github.com/js-cookie/js-cookie/pull/237#issuecomment-232860139) shows how we are testing in node without having to use Mocha.

Gruntfile.js
lynndylanhurley: I'm not sure I understand the problem. Are you saying you want mocha to run without running babel? If we do that, then we can't test es6 imports.

FagnerMartinsBrack: > Are you saying you want mocha to run without running babel? If we do that, then we can't test es6 imports.

No, I am saying to run Babel without mocha, then we might be able to use ES6 imports, just dumping the test runner.

test/es6-import.js
lynndylanhurley: How do you use es6 imports with nodeunit?

FagnerMartinsBrack: That's what we need to look for. We are not using mocha in this project.

lynndylanhurley:
Is there any chance of getting my initial PR merged?

FagnerMartinsBrack:
> Is there any chance of getting my initial PR merged?

You mean this one? Yes, but I can only review it in detail and merge on the weekend (tomorrow AEST).

lynndylanhurley:
Take your time.

FagnerMartinsBrack:
Why closing it?

PR #248, opened 2016-07-31
Update grunt-contrib-uglify to version 2.0.0

FagnerMartinsBrack:


FagnerMartinsBrack:
The order of commits on this PR proves that the problem of failing on those browsers was because of the backwards incompatible changes of the `screwIE8` option. See https://github.com/js-cookie/js-cookie/pull/242#issuecomment-236415359.

Issue #238, opened 2016-07-15
Investigate issues with js-cookie and window.json


FagnerMartinsBrack:
There are some issues being reported about it, we need to investigate why this is leaking and how w can fix it without a huge impact in byte size:

* https://github.com/js-cookie/js-cookie/issues/235
* https://github.com/js-cookie/js-cookie/pull/237#issuecomment-232858028
* https://github.com/js-cookie/js-cookie/issues/117

This can also happen when we have an element with and id `json` instead of just `window.json`.

jshuaf:
Changing `this.json` to `window.json` fixed it for me. Prior to the fix, `Cookies.get()` always returned empty.

PR #250, opened 2016-07-31
Do not parse as JSON when window has a truthy 'json' prop

FagnerMartinsBrack:
Related to #238.

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -127,7 +127,7 @@
 
 		api.set = api;
 		api.get = function (key) {
-			return api(key);
+			return api.call(api, key);
 		};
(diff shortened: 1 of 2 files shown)


src/js.cookie.js
```diff
@@ -127,7 +127,7 @@
 
 		api.set = api;
 		api.get = function (key) {
-			return api(key);
+			return api.call(api, key);
```
ellenaua: Why do you pass "api" to this, if in "this" it expects window

FagnerMartinsBrack: It does not expects `window`. See https://github.com/js-cookie/js-cookie/issues/251#issuecomment-238235648.

Issue #239, opened 2016-07-17
Add docs of how to use it using ES6 module imports


FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/issues/233#issuecomment-233187386

PR #256, opened 2016-08-22
Add docs to use with ES6 imports

FagnerMartinsBrack:
Related to https://github.com/js-cookie/js-cookie/issues/239.

@ThaNarie can you review this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ JavaScript Cookie supports [npm](https://www.npmjs.com/package/js-cookie) and [B
 
 ### Module Loaders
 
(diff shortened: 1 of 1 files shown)


mediamonks-arjan:
@FagnerMartinsBrack I would just call them 'es6' modules (leaving out the 'imports' part) :)

LGTM otherwise!

FagnerMartinsBrack:
@mediamonks-arjan done, can you take a look again?

ThaNarie:
(oops, was using the 'wrong' account :))

Looks good!

Issue #244, opened 2016-07-20
AMD over UMD


cllns:
Thanks for `js-cookie`!

I'm using `js-cookie` with Browserify to build a widget. It works on most sites, but it failed on `nytimes.com`. I figured out was because `nytimes.com` uses AMD (that is, it has the `define` function defined in the global namespace) where as Browserify uses UMD.

My code was breaking because `js-cookie` defined itself via `define`, which I don't use at all.

Here's the current code:
```

if (typeof define === 'function' && define.amd) {
    define(factory);
} else if (typeof exports === 'object') {
    module.exports = factory();
} else {
    var OldCookies = window.Cookies;
    var api = window.Cookies = factory();
    api.noConflict = function () {
        window.Cookies = OldCookies;
        return api;
    };
}

```

I was able to fix it by swapping the order of the checks, so that it checks for UMD (`(typeof exports === 'object'`) before AMD (`(typeof define === 'function' && define.amd)`).

This just creates the opposite problem though, of preferring UMD over AMD.

In order to accommodate both, what about something like:
```
var usesAmdOrUMD = false;
if (typeof define === 'function' && define.amd) {
    usesAmdOrUMD = true;
    define(factory);
}

if (typeof exports === 'object') {
    usesAmdOrUMD = true;
    module.exports = factory();
}

if (!usesAmdOrUMD) {
    var OldCookies = window.Cookies;
    var api = window.Cookies = factory();
    api.noConflict = function () {
        window.Cookies = OldCookies;
        return api;
    };
}

```
This would have the effect of defining it *both* via AMD and UMD, if they're both available. I'm not sure if that's OK or not.

Maybe there's some other way of allowing either without preferring one over the other? Not sure how that'd be possible though.

I think this is the same issue as #173 and #201
And this is possibly related: #234

FagnerMartinsBrack:
Hey, thanks for the explanation and links! Is there any way we can create a test for this first so that we can think of the best way to fix it?

We have a few [AMD tests](https://github.com/js-cookie/js-cookie/blob/960a75ed4f5876327c2a7628cc6e41be282effcc/test/amd.js) already, maybe we could use a similar structure to test this use case adding the minimum amount of dependencies to reproduce the issue?

FagnerMartinsBrack:
@cllns I got some time to read it better and try to understand the problem in details, so let me see if I got the problem right:

The environment uses AMD and has a global `define` function. The environment also uses `UMD` and has a global `exports` object.

[...]

PR #257, opened 2016-08-22
Register in AMD and UMD if both available

FagnerMartinsBrack:
Related to https://github.com/js-cookie/js-cookie/issues/244.

Sadly, this adds +8 bytes gzipped, which breaks the feature of `~800 bytes gzipped` stated [on the README](https://github.com/js-cookie/js-cookie/tree/72d05c27ebfb99aa8b6ae8f1b46210e223e815b9#javascript-cookie--) (not we have `901 bytes`):

screenshot 2016-08-22 13 12 03

@cllns can you please review this?

Also, do you have any idea on how we can make it so that it doesn't add `8 bytes` gzipped in a way that we don't need to update the README? Even reasonably alternative solutions for this use case that don't require changing the library will help not increasing the code size.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -31,6 +31,7 @@ module.exports = function (grunt) {
 					urls: [
 						'http://127.0.0.1:9998/',
 						'http://127.0.0.1:9998/amd.html',
+						'http://127.0.0.1:9998/environment-with-amd-and-umd.html',
 						'http://127.0.0.1:9998/encoding.html?integration_baseurl=http://127.0.0.1:9998/'
 					]
 				}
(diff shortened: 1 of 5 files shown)


cllns:
It's much less clear but what about using boolean expressions instead of the conditionals? We can leverage the bitwise OR (`|`), which will run both sides and *not* short-circuit like `||` would when the first expression is true.

I haven't tested it, but something like:
```
if !(
     (typeof define === 'function' && define.amd && define(factory) ) |
     (typeof exports === 'object' && module.exports = factory() )
) {
    var OldCookies = window.Cookies;
    var api = window.Cookies = factory();
    api.noConflict = function () {
        window.Cookies = OldCookies;
        return api;
    };
}
```

FagnerMartinsBrack:
@cllns Thanks for the suggestion, I tried to make it work but failed to do so. I couldn't find a way to check if `define(factory)` or `module.exports = factory()` was called to decide whether we should create the global `Cookies` or not (`window.Cookies` is not accessible if the environment supports AMD or UMD).

We will probably have to compensate the bytes later by removing semicolons, strict equality, etc. For now I will just document `~900 bytes` gzipped instead of `~800 bytes gzipped`... ¯\\_(ツ)_/¯

Issue #260, opened 2016-09-01
Get back the library down to 800 bytes gzipped


FagnerMartinsBrack:
As from https://github.com/js-cookie/js-cookie/pull/257#issuecomment-242815615, the library size became more than 900 bytes gzipped.

Consider removing semicolons, changing the way IIFE is done and other quirks to see if they are worth it.

FagnerMartinsBrack:
Removing semicolons doesn't work because minifiers need to put it in the minified version to be able to write everything on a single line. Using newline (`\n`) instead would yield no difference because it is 1 byte and could even make the size worse because the minified file doesn't contain any newline (and therefore the gzip algorithm would have to create a new index for storing the newline (`\n`) occurrences).

Changing the IIFE from `(function () {}())` to `!function() {}` doesn't seems to yield any difference either.

Issue #78, opened 2015-08-23
Add support for "First-Party-Only" cookies


FagnerMartinsBrack:
https://tools.ietf.org/html/draft-west-first-party-cookies-03

> The "First-Party-Only" attribute limits the scope of the cookie such
   that it will only be attached to requests if those requests are
   "first-party", as described in Section 2.1.  For example, requests
   for "https://example.com/sekrit-image" will attach first-party-only
   cookies if and only if the top-level browsing context is currently
   displaying a document from "https://example.com".

Better wait for a final version of the draft before deciding how to support it.

FagnerMartinsBrack:
Updated https://tools.ietf.org/html/draft-west-first-party-cookies-04

FagnerMartinsBrack:
Updated https://tools.ietf.org/html/draft-west-first-party-cookies-05

FagnerMartinsBrack:
Updated https://tools.ietf.org/html/draft-west-first-party-cookies-06

FagnerMartinsBrack:
It is closer to 1 year since I found out about this spec. It doesn't seem to be worth keeping an open issue for this, let's address once there is a legit use case or request for that.

PR #264, opened 2016-09-06
Support for same-site cookie attribute

kukulich:


FagnerMartinsBrack:
Thanks for the PR but we need a few pieces of information first:

Is that standardized already, if not what is the browser support? I didn't find anything on [RFC 6265](https://tools.ietf.org/html/rfc6265), which is the official cookie spec.

Does it have any relationship with https://github.com/js-cookie/js-cookie/issues/78?

kukulich:
More info: http://www.sjoerdlangkemper.nl/2016/04/14/preventing-csrf-with-samesite-cookie-attribute/
Browser support: https://www.chromestatus.com/feature/4672634709082112
Specification: https://tools.ietf.org/html/draft-west-first-party-cookies-07

I hope this is enough :)

FagnerMartinsBrack:
@kukulich Thanks for the links.

It seems based on the browser support that the feature is not official yet and we don't support unofficial features (see #78 for an example).

Can you tell your actual use case for that just for context?

Also, what we could do is create a hook so that the developer can decide which attributes they want to support without having to change the library. Maybe allowing that any attribute that is passed in the last argument of the `Cookies.set()` method is directly added to the cookie without us having to previously support it? Does that make sense?

kukulich:
The use case is easy. I just want to protect all of my cookies against CSRF. We create some cookies in javascript so I want to add SameSite attribute to them.

I'm ok with your suggestion, it will work for us too.

FagnerMartinsBrack:
@kukulich I have received a notification you added a new commit to this Pull Request, probably because you opened it from `master` branch. Are you still tracking this? If not I guess this can be closed.

kukulich:
@FagnerMartinsBrack 

I'm sorry, pull request is fixed.

I still hope js-cookie will somehow support SameSite. I only waited for your response. Should I try to prepare pull request based on your suggestion?

FagnerMartinsBrack:
> Should I try to prepare pull request based on your suggestion?

You can, then we can discuss the feasibility there. Discussion with code is always better :D

kukulich:
@FagnerMartinsBrack Pull request updated.

FagnerMartinsBrack:
Is it possible for us to create a code that gzipped and minified accounts for less (or at least as closer as possible to) 900 bytes? We have this "900 bytes gzipped" feature and [a task to get it down to 800 again](https://github.com/js-cookie/js-cookie/issues/260).

kukulich:
I will look at it.

kukulich:
Gzipped and minified code is 998 bytes now. I don't have any idea how to shorter this code. However the official support of SameSite attribute would be shorter for sure, it would be only 917 bytes:

```diff
- attributes.secure ? '; secure' : ''
+ attributes.secure ? '; secure' : '',
+ attributes.sameSite ? '; SameSite=' + attributes.sameSite : ''
```

FagnerMartinsBrack:
As stated in [the comment above](https://github.com/js-cookie/js-cookie/pull/264#issuecomment-245195599), we don't support unofficial features. If making the attributes API more free will add that many bytes, I think it's not worth doing that right now.

I realized same-site cookie is the same a first-party cookies, which is a duplicate of https://github.com/js-cookie/js-cookie/issues/78.

[The draft](https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00) was reviewed recently and at the moment of writing it was re-marked to expire at December 22, 2016. When [the draft stops being a draft](https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00) and updates RFC 6265, then I would be very keen on adding it as a supported feature with proper documentation too. Before that it is a risk because they can just drop it.

For now, I don't think it's worth it, but feel free to fork the project and add the feature in a per use case basis. Eventually, it may get it's way to js-cookie, unless they drop the attribute from the spec.

FagnerMartinsBrack:
@kukulich Given that somebody created https://github.com/js-cookie/js-cookie/pull/275#issue-183656360 and considering the comment https://github.com/js-cookie/js-cookie/issues/276#issuecomment-254838817 I am willing to keep reviewing this with the intent to land it before the RFC is finalized.

I believe we can make this code shorter in bytes by removing the `attributes.hasOwnProperty` instances and the `typeof` checks. We don't need to validate user input since the docs are what defines what the input accepts and what not (I have written [an article](https://medium.com/@fagnerbrack/library-internals-do-not-exist-1ff50a30e3b0) explaining about this sort of contract).

Since you are the first one who came up with this, would you be willing to put some effort in adjusting this PR to reduce the number of bytes and add a good test coverage for the feature of allowing any attribute to be set to the cookie?

kukulich:
@FagnerMartinsBrack Could you please open this pull request?

FagnerMartinsBrack:
@kukulich I can't, the reopen button is disabled. Have you removed the original branch? Maybe there's a time limit to reopen =/

I guess we need to create a new PR. You can put a cross link to reference this one, though.

kukulich:
New pull request: https://github.com/js-cookie/js-cookie/pull/284

PR #273, opened 2016-10-03
Add more information about #remove method

Bartuz:
It took me few minutes to find it on my own. Hope to save time of other developers! 

Just a useful info about remove method :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,17 @@ Delete cookie:
 Cookies.remove('name');
 ```
 
+Please notice that removing unexisting cookie does not raise any exception or value
+
+```javascript
+var whatWasTheValue = Cookies.remove('nothing');
+if (whatWasTheValue === undefined) {
+  console.log("I will be excuted");
+} else {
+  console.log("I will never be executed");
+}
+```
+
 Delete a cookie valid to the path of the current page:
 
 ```javascript
--- a/README.md
+++ b/README.md
@@ -87,9 +87,9 @@ Cookies.remove('name');
 Please notice that removing unexisting cookie does not raise any exception nor return any value
 
 ```javascript
-var whatWasTheValue = Cookies.remove('nothing');
-if (whatWasTheValue === undefined) {
-  console.log("I will be excuted");
+var cookieValue = Cookies.remove('nothing');
+if (cookieValue === undefined) {
+  console.log("I will be executed");
 } else {
   console.log("I will never be executed");
 }
--- a/README.md
+++ b/README.md
@@ -84,10 +84,10 @@ Delete cookie:
 Cookies.remove('name');
 ```
 
-Please notice that removing unexisting cookie does not raise any exception or value
+Please notice that removing unexisting cookie does not raise any exception nor return any value
 
 ```javascript
-var whatWasTheValue = Cookies.remove('nothing');
+var  = Cookies.remove('nothing');
 if (whatWasTheValue === undefined) {
   console.log("I will be excuted");
 } else {
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Cookies.remove('name');
 Please notice that removing unexisting cookie does not raise any exception nor return any value
 
 ```javascript
-var  = Cookies.remove('nothing');
+var whatWasTheValue = Cookies.remove('nothing');
 if (whatWasTheValue === undefined) {
   console.log("I will be excuted");
 } else {


README.md
```diff
@@ -84,6 +84,17 @@ Delete cookie:
 Cookies.remove('name');
 ```
 
+Please notice that removing unexisting cookie does not raise any exception nor return any value
```
FagnerMartinsBrack: That doesn't mean the cookie doesn't exist, that means the cookie is not accessible. It is impossible to verify if the cookie exist or not.

We have discussed this in the past, see https://github.com/carhartl/jquery-cookie/issues/164#issuecomment-19003359.

README.md
```diff
@@ -84,6 +84,17 @@ Delete cookie:
...
+
+```javascript
+var whatWasTheValue = Cookies.remove('nothing');
+if (whatWasTheValue === undefined) {
+  console.log("I will be excuted");
```
FagnerMartinsBrack: Typo

README.md
FagnerMartinsBrack: That doesn't mean the cookie doesn't exist, that means the cookie is not accessible. It is impossible to verify if the cookie exist or not.

We have discussed this in the past, see https://github.com/carhartl/jquery-cookie/issues/164#issuecomment-19003359.

README.md
FagnerMartinsBrack: Typo

README.md
```diff
@@ -84,6 +84,17 @@ Delete cookie:
 Cookies.remove('name');
 ```
 
+Please notice that removing unexisting cookie does not raise any exception nor return any value
+
+```javascript
+var whatWasTheValue = Cookies.remove('nothing');
```
FagnerMartinsBrack: We can name the variable consistently throughout the docs

README.md
```diff
@@ -84,6 +84,17 @@ Delete cookie:
...
+if (whatWasTheValue === undefined) {
+  console.log("I will be excuted");
+} else {
+  console.log("I will never be executed");
```
FagnerMartinsBrack: Maybe a comment instead of `console.log`?

FagnerMartinsBrack:
Maybe we can have a generic documentation saying that it will never throw an exception except for the cases it will (like the `getJSON()` when it misses global `JSON` API, for example)?

README.md
Bartuz: I mainly started this `#remove` research because comment here:
```
Cookies.remove('name'); // fail!
```
didn't make me sure enough if fail means exception, undefined or something else. 

In Ruby `fails` means throw exception, so I needed to double check this one 😆

FagnerMartinsBrack: Makes sense your `// fail!` example.

Would this work then:

```
Cookies.remove('name'); // doesn't work
```

FagnerMartinsBrack: Or maybe:

```
Cookies.remove('name'); // doesn't do anything
```

Or:

```
Cookies.remove('name'); // noop
```

The last one can have accessibility problems because not everyone knows what the `noop` concept means.

The reason why I am suggesting those approaches is because it might be interesting to focus on the least amount of text that fixes any confusion. The README is in a state where any new addition will increase the friction because there will be more content, so it might be worth putting some effort to not add new paragraphs unless essential.

Since you just started with the library you don't have the maintainers bias on the README because we already know how the API works and we can't make it better. Your feedback is really really valuable! Thanks.

Bartuz: It's really nice you appreciate my voice here, thanks :)

I like your approach to keep guides as short as possible but IMHO your new comments may confuse people even more than `fail`. 

The problem is that current `fail` comment is used for the first time in context of `"valid path of current page"`  so even adding super descriptive comment there is not intuitive enough to communicate that even removing Cookie with proper "path" argument also "fails"

I would just add block that every .remove call ends up with undefined, successful or not

```
Cookies.remove('name'); // returns undefined
Cookies.remove('nothing'); // also returns undefined
```

FagnerMartinsBrack: I understand. What about something like this then:

```
Delete a cookie that is visible on the current page:

Cookies.set('name', 'value', { path: '' });
Cookies.remove('name'); // does not remove the cookie
Cookies.remove('name', { path: '' }); // removes the cookie
```

We can then consistently use "visible" instead of "valid" where we are talking about the cookie visibility based on the attributes.

Unfortunately, we can't document that `Cookies.remove()` returns `undefined` because that's not a supported behavior. It is actually `void` and returns nothing.

Make sense?

FagnerMartinsBrack:
> Maybe we can have a generic documentation saying that it will never throw an exception except for the cases it will (like the `getJSON()` when it misses global `JSON` API, for example)?

It seems my previous comment was wrong. We don't throw any exception when `window.JSON` is not present since the code is in a try/catch block.

Actually, it looks like we never throw any exception for documented functionality, maybe we can make that clear in the docs instead of just documenting a single API?

@Bartuz Do you still want to work on this?

FagnerMartinsBrack:
Ping @Bartuz, do you still want to work on this PR?

PR #275, opened 2016-10-18
Added SameSite to cookie via attributes

TBoonX:
I added samesite as a new options attribute. It will either be Strict or Lax. The inputs could be true, 'Strict' or 'Lax'.
For this I also added QUnit tests.

I know this attribute is not in RFC 6265, but this is a minor change which results in a new feature.
We from [SlideWiki](//github.com/slidewiki) will heavily benefit from this one.

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -56,6 +56,13 @@
 					attributes.expires = expires;
 				}
 
+        if (attributes.samesite !== 'Strict' && attributes.samesite !== 'Lax') {
+          if (attributes.samesite === true)
+            attributes.samesite = 'Strict';
+          else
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
Duplicated of https://github.com/js-cookie/js-cookie/issues/78 and https://github.com/js-cookie/js-cookie/pull/264.

My only concern is that it is a draft that is still not approved. It might be worth using a fork for now instead of making a decision to support it for a wider audience. Once it is approved, we would definitely support it.

FagnerMartinsBrack:
Closing in favor of https://github.com/js-cookie/js-cookie/pull/284.

Issue #189, opened 2016-04-29
False advertising?


harrync:
'Very good wiki' - there are 3 pages, i.e. not very good. Might want to remove this sentence?

FagnerMartinsBrack:
> ... there are 3 pages

That is the reason why the wiki is very good, because there is only information that was proven to be essential. Different from other wikis, where people are bombarded with useless information, js-cookie wiki is based on [empirical evidence of useful content or frequently asked questions](https://github.com/js-cookie/js-cookie/issues/37) that will serve the majority of the library users.

Saying that a wiki is not good because it has a few pages is a falacy, because the argument is against the amount of information, not against the quality or usefulness of the content. A wiki can also have a lot of content and be useless or have a low amount of content and be useful.

That is what "very good" is all about in this context. Quality, not quantity. Therefore there is no "false advertising" here.

> Might want to remove this sentence?

I will be glad to do it if you provide some concrete evidence of why the wiki is not very good with arguments against the content, not against the amount of information there is.

Thanks for the feedback.

FagnerMartinsBrack:
@harrync, what about: "Useful Wiki"?

shackra:
I was expecting something alone the lines @harrync so I agree to change the wording to "Useful Wiki" or "Wiki with just the useful information".

PR #284, opened 2016-12-04
Unknown attributes are always added to cookie, eg. SameSite attribute can be set now

kukulich:


kukulich:
```
   raw     gz Compared to master @ 7fd36e66b60a4e0ac9d09207fb3a348a2969152c
     =      = build/js.cookie-2.1.3.min.js
  +376   +115 src/js.cookie.js
```

test/tests.js
```diff
@@ -241,8 +241,29 @@ QUnit.test('false secure value', function (assert) {
...
+	assert.expect(1);
+	assert.strictEqual(Cookies.set('c', 'v', {
+		unofficial: 'anything'
+	}), 'c=v; path=/; unofficial=anything', 'should write the cookie string with unofficial attribute');
+});
+
+QUnit.test('unofficial attribute with boolean value', function (assert) {
+	assert.expect(1);
+	assert.strictEqual(Cookies.set('c', 'v', {
+		unofficial: true
```
FagnerMartinsBrack: Don't we need to allow only `SameSite=strict` or `SameSite=lax`? Do we need to allow `SameSite`? What is the difference?

src/js.cookie.js
```diff
@@ -74,13 +81,21 @@
...
 					key, '=', value,
 					attributes.expires ? '; expires=' + attributes.expires.toUTCString() : '', // use expires attribute, max-age is not supported by IE
 					attributes.path ? '; path=' + attributes.path : '',
 					attributes.domain ? '; domain=' + attributes.domain : '',
 					attributes.secure ? '; secure' : ''
-				].join(''));
+				];
+
+				for (var attribute in attributes) {
+					if (!officialAttributes.hasOwnProperty(attribute) && attributes[attribute]) {
```
FagnerMartinsBrack: Why are we using `officialAttributes.hasOwnProperty(attribute)`.

src/js.cookie.js
```diff
@@ -24,6 +24,13 @@
 		};
 	}
 }(function () {
+	var officialAttributes = {
```
FagnerMartinsBrack: I understand that the intent here is to put the official attributes in the top do document what is official and what is not. In this case, though, we can try to reduce the number of bytes shifting this around since we already have tests and the public API docs that can document what is officially supported and what not.

It's ok to trade clean code for the sake of code size in this project.

test/tests.js
```diff
@@ -274,6 +298,13 @@ QUnit.test('with attributes', function (assert) {
 	assert.strictEqual(document.cookie, '', 'should delete the cookie');
 });
 
+QUnit.test('deletion with unofficial attributes', function (assert) {
+	assert.expect(1);
+	var attributes = { path: '/', unofficial: 'anything' };
```
FagnerMartinsBrack: Do we actually need to pass the unofficial attributes to be able to remove the cookie or was this just following the previous examples?

FagnerMartinsBrack:
>    raw     gz Compared to master @ 7fd36e66b60a4e0ac9d09207fb3a348a2969152c
>     =      = build/js.cookie-2.1.3.min.js
>  +376   +115 src/js.cookie.js

We probably need to document the byte increase in the README to prevent [this kind of thing](https://github.com/js-cookie/js-cookie/issues/189).

FagnerMartinsBrack:
I have made a few comments, let's see how far we can tweak this before landing to `master`. After we statically review, I will try to checkout locally and see what else we can do to remove bytes and post the suggestions here.

Thanks!

src/js.cookie.js
kukulich: I just need the list of official attributes somewhere. And I've found a place that is called just once. There's no other intent :)

kukulich:
Actual version:

```
   raw     gz Compared to master @ 7fd36e66b60a4e0ac9d09207fb3a348a2969152c
     =      = build/js.cookie-2.1.3.min.js
  +361   +100 src/js.cookie.js
```

src/js.cookie.js
FagnerMartinsBrack: Oh ok, I thought it was for cleanliness purposes :D.

test/tests.js
FagnerMartinsBrack: We need to make sure it's required, otherwise we will create code to support speculation and that's not good =(. We can either remove the test and leave it as an unsupported feature until we get a request or research how is the browser behavior and document that.

Since we are not "officially" supporting custom attributes I don't think we need to make the research. However, if we are going to document that we support it, then we definitely need to make sure what we are doing.

test/tests.js
FagnerMartinsBrack: I understand. Let's review the focus of this PR then:

1. Allow the `SameSite` attribute to be passed for those willing to use it at their own risk. js-cookie won't support it explicitly so we won't document how it works, until the attribute is amended to the official RFC 6265 spec.
2. Allow the same code to be generic in a way that allows people to experiment using custom attributes that are different than the `SameSite` attribute.
3. Reduce the amount of code we need to add in js-cookie and the amount of behavior we need to support

For the number **1** we can just add support for the `{ unnoficial: String }` structure. For the number **2** we need to add `{ unnoficial: true }`, however that will conflict with number **3**.

What if we build number **1** and leave number **2** out for now in a way that we can achieve number **3** without any issue? Later we can implement number **2** if there's enough requests.

Thoughts?

test/tests.js
FagnerMartinsBrack: See https://github.com/js-cookie/js-cookie/pull/284#discussion_r90991200

FagnerMartinsBrack: > However you need to pass all official attributes I think, so I prefer to do it the same way.

About that, I don't think we need to pass all official attributes. According [to the docs](https://github.com/js-cookie/js-cookie/tree/7fd36e66b60a4e0ac9d09207fb3a348a2969152c#basic-usage), it's just the "exact same path and domain attributes that was used to set the cookie".

It would be nice to confirm if we need the "secure" flag too and document that!

FagnerMartinsBrack:
Does it make sense changing the README to add the new byte size in the ["features" listing](https://github.com/js-cookie/js-cookie/tree/7fd36e66b60a4e0ac9d09207fb3a348a2969152c#javascript-cookie--) and update it as we manage to reduce the size? This way we won't forget to do it later because the diff will highlight the changes.

test/tests.js
kukulich: Without boolean support:
```
   raw     gz Compared to master @ 7fd36e66b60a4e0ac9d09207fb3a348a2969152c
  +106    +61 build/js.cookie-2.1.3.min.js
  +321    +90 src/js.cookie.js
```

So it's only 9-10 bytes less. Do you thing it is worth?

test/tests.js
kukulich: Chrome can delete cookie without the `SameSite` flag so I've removed the test.

test/tests.js
FagnerMartinsBrack: > So it's only 9-10 bytes less. Do you thing it is worth?

Yes, even 1 byte is relevant because we're talking about GZipped size. This library is used by a few big websites and at scale it can make a difference.

FagnerMartinsBrack: Another good thing about cutting scope is that the PR becomes simpler so it's easy to focus on what really matters, which is the `SameSite` for now.

kukulich:
Updated.

FagnerMartinsBrack:
I have [made some changes](https://github.com/slevomat/js-cookie/pull/1). You'll retain full authorship despite that.

FagnerMartinsBrack:
Ping @kukulich, are you still looking into this?

kukulich:
Your code looks good to me :)

FagnerMartinsBrack:
Closing in favor of https://github.com/js-cookie/js-cookie/pull/293

PR #288, opened 2016-12-13
Use strict mode, avoid octal syntax in tests.js

ysangkok:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode , strict mode does not tolerate leading zeroes. This was not meant to be octal anyway, so thank God the number was not bigger than 7. When I run this through Babel, I get the SyntaxError. It is therefore not purely theoretical.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -60,7 +60,7 @@ module.exports = function (grunt) {
 				requireSpaceBeforeObjectValues: true,
 				validateIndentation: '\t',
 				validateLineBreaks: 'LF',
-				validateQuoteMarks: true,
+				validateQuoteMarks: {mark: '\'', escape: false},
 				disallowSpacesInsideArrayBrackets: 'all',
 				disallowSpacesInsideParentheses: true,
 				disallowTrailingWhitespace: true
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -3,6 +3,7 @@
 	"qunit": true,
 
 	"-W053": true,
+	"-W097": true, // disable warning for 'Use the function form of "use strict".'
 
 	"extends": "../.jshintrc",
 	"globals": {
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -3,7 +3,7 @@
 	"qunit": true,
 
 	"-W053": true,
-	"-W097": true, // disable warning for 'Use the function form of "use strict".'
+	"-W097": true,
 
 	"extends": "../.jshintrc",
 	"globals": {
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -3,7 +3,7 @@
 	"qunit": true,
 
 	"-W053": true,
-	"-W097": true,
+	"-W097": true, // disable warning for 'Use the function form of "use strict".'
 
 	"extends": "../.jshintrc",
 	"globals": {
--- a/test/tests.js
+++ b/test/tests.js
@@ -1,4 +1,4 @@
-"use strict";
+'use strict';
 
 /*global lifecycle: true*/
 
--- a/test/tests.js
+++ b/test/tests.js
@@ -404,7 +404,7 @@ QUnit.test('Object Constructor', function (assert) {
 
 QUnit.test('Use String(value) for unsupported objects that do not stringify into JSON', function (assert) {
 	assert.expect(2);
-	Cookies.set('date', new Date(2015, 04, 13, 0, 0, 0, 0));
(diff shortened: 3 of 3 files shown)


FagnerMartinsBrack:
@ysangkok Can we run the tests in "strict mode" so that the `04` Number usage throws the `Octal literals are not allowed in strict mode` syntax error?

ysangkok:
@FagnerMartinsBrack Yes, I added the 'use strict' pragma.

FagnerMartinsBrack:
Build is failing.

I suggest to remove the octal fix and make the code fail first for a reason we expect (not let it fail because of jshint, but because of the octal), then add the "use strict" later to make it green again.

This way we can be sure that the code will do what it should and there will be no unexpected build breakage in the PR.

ysangkok:
@FagnerMartinsBrack , I could wrap the whole file in a function, but it would be ugly, and would involve changing all lines to satisfy jscs. I don't know any other way...

FagnerMartinsBrack:
@ysangkok We can try to ignore that specific rule from jshint in test files. See the [.jshintrc file](https://github.com/js-cookie/js-cookie/blob/b1f0a45ac82c892442215fd3cc54877465f3ae8b/test/.jshintrc).

ysangkok:
@FagnerMartinsBrack OK, now the tests should fail because of the octal value. Do you want a new pull request for this? Do you want a separate pull request for the fix?

FagnerMartinsBrack:
@ysangkok Why a new Pull Request? Can't we just ignore the rule of using the function form of "use strict" in the test folder?

ysangkok:
@FagnerMartinsBrack ok tell me if this is all right.

test/.jshintrc
```diff
@@ -3,6 +3,7 @@
 	"qunit": true,
 
 	"-W053": true,
+	"-W097": true, // disable warning for 'Use the function form of "use strict".'
```
FagnerMartinsBrack: JSON spec doesn't allow comments.

ysangkok: According to https://github.com/jshint/jshint/issues/741 , the file can contain comments. Should I still remove it?

ysangkok:
I added a commit configuring the quote style so that it is clear that single quotes are preferred. Still waiting on your feedback regarding the comment in jshintrc.

test/.jshintrc
FagnerMartinsBrack: See https://github.com/jshint/jshint/issues/741#issuecomment-268786491. If they explicitly support it, so let's keep it. However, it doesn't seem wise to rely on the word of mouth to create exception for the format, unless it's explicitly documented.

Gruntfile.js
```diff
@@ -60,7 +60,7 @@ module.exports = function (grunt) {
 				requireSpaceBeforeObjectValues: true,
 				validateIndentation: '\t',
 				validateLineBreaks: 'LF',
-				validateQuoteMarks: true,
+				validateQuoteMarks: {mark: '\'', escape: false},
```
FagnerMartinsBrack: Oh, I see. Because they encountered "use strict" they consider the whole file should use double quotes. Wouldn't be easier to use `'use strict'` with single quotes then? If we do that then we probably don't need to change the jshint rule.

FagnerMartinsBrack: Unless I am missing something?

ysangkok: I changed it to use single quotes, but I figured it would be better to explicitly state the style of quotes used. You do not agree?

FagnerMartinsBrack: I don't think `validateQuoteMarks: {mark: '\'', escape: false}` is any different of `validateQuoteMarks: true`. I understand the [Conascence](https://en.wikipedia.org/wiki/Connascence_(computer_programming)) is big for the `validateQuoteMarks: true` and we should probably improve it. However, if such change is not necessary right now, it can be discussed in a separate PR since this one is just about the octal issue.

test/.jshintrc
```diff
@@ -3,7 +3,7 @@
 	"qunit": true,
 
 	"-W053": true,
-	"-W097": true, // disable warning for 'Use the function form of "use strict".'
```
FagnerMartinsBrack: We can also create a new PR that adds the comments in all cryptic parameters depending on the response of https://github.com/jshint/jshint/issues/741#issuecomment-268786491

ysangkok:
Ok, is something left to be done?

FagnerMartinsBrack:
@ysangkok all good, thanks.

PR #291, opened 2016-12-20
Minor usability improvement to the readme.md

a2wd:
Added an anchor in the converters section heading for the link from the encodings section to point to (previously, the anchor tag/name attribute didn't exist so the link didn't do anything)

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Cookies.getJSON(); // => { name: { foo: 'bar' } }
 
 This project is [RFC 6265](http://tools.ietf.org/html/rfc6265#section-4.1.1) compliant. All special characters that are not allowed in the cookie-name or cookie-value are encoded with each one's UTF-8 Hex equivalent using [percent-encoding](http://en.wikipedia.org/wiki/Percent-encoding).  
 The only character in cookie-name or cookie-value that is allowed and still encoded is the percent `%` character, it is escaped in order to interpret percent input as literal.  
-Please note that the default encoding/decoding strategy is meant to be interoperable [only between cookies that are read/written by js-cookie](https://github.com/js-cookie/js-cookie/pull/200#discussion_r63270778). To override the default encoding/decoding strategy you need to use a [converter](#converter).
+Please note that the default encoding/decoding strategy is meant to be interoperable [only between cookies that are read/written by js-cookie](https://github.com/js-cookie/js-cookie/pull/200#discussion_r63270778). To override the default encoding/decoding strategy you need to use a [converter](#converters).
 
 ## Cookie Attributes
 
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ Cookies.get('name'); // => 'value'
 Cookies.remove('name', { secure: true });
 ```
 
-## Converters <a name="converter"></a>
+## Converters
 
 ### Read
 


FagnerMartinsBrack:
I don't understand what this do. All headers in Github markdown have links Can you please elaborate or record a video showing the difference? Thanks.

FagnerMartinsBrack:
Closing due to the lack of feedback from the [OP](https://medium.com/@fagnerbrack/what-does-op-means-on-github-8db958b604af). If you want to continue the discussion, please comment here mentioning one of the members.

a2wd:
@FagnerMartinsBrack sorry for the delay, I've added a screengrab (below) unless it gets compressed, full size you should be able to see the difference - the left is the version in this PR, the right if the current master.

comparison

FagnerMartinsBrack:
Shouldn't it be `#converters` with an `s` in the end? See https://github.com/js-cookie/js-cookie/tree/b1f0a45ac82c892442215fd3cc54877465f3ae8b#converters

FagnerMartinsBrack:
Probably the fix is in the command, not the header. The link should link to `#converters` instead.

FagnerMartinsBrack:
It looks like this typo has existed [since June 2016 when it was added](https://github.com/js-cookie/js-cookie/commit/960a75ed4f5876327c2a7628cc6e41be282effcc). See the "Encoding" section [in the context of that revision](https://github.com/js-cookie/js-cookie/blob/960a75ed4f5876327c2a7628cc6e41be282effcc/README.md#encoding).

a2wd:
In that revision and previous ones I looked at it doesn't appear to have an anchor or name. The github markdown generator does add a custom `#user-content-convertors` anchor nearby the heading though, but presumably that's not what you want to link to?

FagnerMartinsBrack:
> In that revision and previous ones I looked at it doesn't appear to have an anchor or name.

Maybe Github does some magic, the important thing is that it works even without an `id` in the header element:

screenshot 2017-01-09 21 32 56

(Don't ask me why)

> The github markdown generator does add a custom #user-content-convertors anchor nearby the heading though, but presumably that's not what you want to link to?

That's not what we should link to. That's Github generated markup. We shouldn't be adding to the project's Markdown level something that is of knowledge of Github only. Github can change the `id="user-content-converters"` at any time, the raw HTML is strictly internal behavior of Github, and it has the same problem as when we [rely on library internals](https://medium.com/@fagnerbrack/library-internals-do-not-exist-1ff50a30e3b0).

In this case, the best approach is to simulate the same link that Github adds when you click on the anchor in the left side of the title, see the image below:

screenshot 2017-01-09 21 26 16

That link points to https://github.com/js-cookie/js-cookie/tree/4e77d05b59692b664fe869f3e2dd54b77c13badf#converters

The likelihood of Github supporting their header generated anchor format is much higher than supporting a consistent generated HTML markup. Since they don't explicitly document that anchor and how link to them (or at least I'm not aware of any docs about it) this is still internal, but it's using an internal that if changed will also break several other projects out there. For that reason, it's safer than inspecting the HTML markup and linking to a custom anchor tag like `#user-content-converters`.

The `<a name="converter"></a>` is also a good approach because it ensures we have control of the element id. However, it mixes HTML with markdown but no apparent benefit while also breaking consistency with the rest of the file.

Given the tradeoffs, fixing the [line 145](https://github.com/js-cookie/js-cookie/blame/4e77d05b59692b664fe869f3e2dd54b77c13badf/README.md#L145) to point to `[converter](#converters)` seems to be the best approach in this case.

Make sense?

a2wd:
@FagnerMartinsBrack as discussed, I attempted with a change to #converters but that doesn't help either - you can see this at https://github.com/a2wd/js-cookie/blob/master/README.md#converters

FagnerMartinsBrack:
> @FagnerMartinsBrack as discussed, I attempted with a change to #converters but that doesn't help either - you can see this at https://github.com/a2wd/js-cookie/blob/master/README.md#converters

It doesn't work with your link because you have changes in the header where there's an additional space on it before the tag you have added in the markdown, so the header anchor is interpreted as `#converters-` instead of `#converters`. Here's the same link you posted but now working: https://github.com/a2wd/js-cookie/blob/master/README.md#converters-

By the way, according to the [contributing guidelines](https://github.com/js-cookie/js-cookie/blob/master/CONTRIBUTING.md#pull-requests), we require PRs to be open using a topic/feature branch instead of `master`. See this for additional info: https://help.github.com/articles/about-pull-requests/

README.md
```diff
@@ -225,7 +225,7 @@ Cookies.get('name'); // => 'value'
 Cookies.remove('name', { secure: true });
 ```
 
-## Converters
+## Converters <a name="converter"></a>
```
FagnerMartinsBrack: As explained in the [submitted review](https://github.com/js-cookie/js-cookie/pull/291#pullrequestreview-17196201), if you remove `<a name="converter"></a>`, then the link will just work.

FagnerMartinsBrack:
Ping @a2wd. Are you still looking into this?

a2wd:
Again, sorry for the delay - this is now done and working 👍

PR #298, opened 2017-02-02
Add .has method

th3fallen:
Add a .has method to easily determine if a cookie exists

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,13 @@ Read all visible cookies:
 Cookies.get(); // => { name: 'value' }
 ```
 
+Check if cookie exists:
+
+```javascript
+Cookies.has('name'); // => true
+Cookies.has('nothing'); // => false
+```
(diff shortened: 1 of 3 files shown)


FagnerMartinsBrack:
@th3fallen It's impossible to verify a cookie exist because it can be created with a different path and domain options. If it's on a different path or domain it exists but it can't be accessed, so it will always provide the false negative.

When this project used the jquery namespace we used to return `true` or `false` if the cookie was removed or not through `$.removeCookie()`. It turns out it's impossible to efficiently determine if the cookie exists or not.

For context, please see https://github.com/carhartl/jquery-cookie/issues/382#issuecomment-166916510, https://github.com/carhartl/jquery-cookie/issues/376#issue-111944758, https://github.com/carhartl/jquery-cookie/issues/336#issuecomment-72659244

Issue #270, opened 2016-09-27
Default value for cookies


james4388:
When I need to provide default value to cookie key I have to do:
```
var value = Cookies.get('key');
if (typeof(value) === 'undefined') {
    value = default_value;
}
```

Can we modify api to accept default value? Is there any reason not to do so? May I provide a pull request?

FagnerMartinsBrack:
We had similar discussions in the past, see https://github.com/js-cookie/js-cookie/issues/202 and https://github.com/js-cookie/js-cookie/issues/135.

james4388:
Well you can do Cookies.get('key') || 42 and it return 42 but thing come complected when cookie value is true / false
Cookies.set('key', false)
Cookies.get('key') || true => true
But since all cookie value become string then this problem is not an issue. You can close this then.

PS to preserve get parameter for something useful in the future. we can use arguments.

FagnerMartinsBrack:
The only issue could be when the value is an empty String, but in that case, it is easy to create a wrapper that does something like:

```javascript
getCookieWithDefaultValue('key', '42');
```

And uses js-cookie underneath.

Issue #202, opened 2016-05-20
Support default value for get method


hurelhuyag:
Idea is if cookie is not available then return 'defaultvalue'

Cookies.get('cookiename', 'defaultvalue');

FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/issues/135 for a whole discussion on this.

What is the difference between `Cookies.get('cookiename', 'defaultvalue')` and `Cookies.get('cookiename') || 'defaultvalue'`?

FagnerMartinsBrack:
Closing due to the lack of feedback from the [OP](https://medium.com/@fagnerbrack/what-does-op-means-on-github-8db958b604af). If you want to continue the discussion, please comment here mentioning one of the members.

Issue #135, opened 2016-02-10
Default value on get


Soullivaneuh:
To avoid conditional code and get it proper, I suggest you to implement default value on `Cookies.get` and `Cookies.getJson`.

Examples (we assume the cookie does not exists):

```javascript
Cookies.get('foo') // undefined
Cookies.get('bar', 42) // 42
```

FagnerMartinsBrack:
What is the practical difference between these two?

```javascript
Cookies.get('bar', 42); // => 42
```

and

```javascript
Cookies.get('bar') || 42; // => 42
```

If we build something as simple as this in the `get` API we would prevent the second argument to be extended for something more useful in the future. There is no sense in changing an API to do something that can be easily achieved with one line.

Soullivaneuh:
Indeed, I didn't know about this trick.

FagnerMartinsBrack:
Ok, closing then. Thanks for the issue :+1:

PR #316, opened 2017-03-29
Add default value to get function

vzool:
Hi,
I modified get function to return _defaultValue_ when value not exists, I made that to provide the functionality for my current project needs without extra **if block**, so I think it is nice to share it with community too.
Thanks for the great job

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Read cookie:
 ```javascript
 Cookies.get('name'); // => 'value'
 Cookies.get('nothing'); // => undefined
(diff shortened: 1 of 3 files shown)


FagnerMartinsBrack:
Why change the library if you can just do:

```javascript
var cookieValue = Cookies.get('nothing') || 'nice';
```

This has been discussed before, see https://github.com/js-cookie/js-cookie/issues/270, https://github.com/js-cookie/js-cookie/issues/202 and https://github.com/js-cookie/js-cookie/issues/135

vzool:
As you like <3

PR #362, opened 2017-09-20
Made the remove() return the removed value

mmcev106:
A minor improvement for convenience in the case where you want to get a cookie then immediately remove it.

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -151,9 +151,13 @@
 		api.defaults = {};
 
 		api.remove = function (key, attributes) {
+			var value = api.get(key);
+
 			api(key, '', extend(attributes, {
 				expires: -1
 			}));
+
+			return value;
 		};
 
 		api.withConverter = init;
--- a/test/tests.js
+++ b/test/tests.js
@@ -297,10 +297,12 @@ QUnit.test('undefined attribute value', function (assert) {
 QUnit.module('remove', lifecycle);
 
 QUnit.test('deletion', function (assert) {
-	assert.expect(1);
-	Cookies.set('c', 'v');
-	Cookies.remove('c');
+	assert.expect(2);
+	var value = 'v';
+	Cookies.set('c', value);
+	var removedValue = Cookies.remove('c');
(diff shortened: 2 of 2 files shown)


src/js.cookie.js
```diff
@@ -151,9 +151,13 @@
 		api.defaults = {};
 
 		api.remove = function (key, attributes) {
+			var value = api.get(key);
+
 			api(key, '', extend(attributes, {
 				expires: -1
 			}));
+
+			return value;
```
FagnerMartinsBrack: Can't you just write:

```javascript
Cookies.remove('mycookie', { path: '/home' });
Cookies.get('mycookie');
```

It makes your code clear that you're trying to retrieve the cookie.

Besides, as you can see in the example, removing a cookie does not guarantee that you have really removed it. You might have called `.remove('mycookie', { path: '/home' })` for a cookie from a different path and you're just not in that path when you call `.get('mycookie')` to it.

mmcev106: I simply thought one line was simpler than two.  I believe many "remove()" like patterns in other libraries and languages return the removed value for just that reason.

FagnerMartinsBrack: With cookies is not that simple because it depends on

1. which path you're in.
2. which domain you're in.

If we return the cookie value, people will start to create bug reports of why it's returning `undefined`, but the problem is most of the times one of the above.

This happened a few years ago when we decided to return `true` for when the cookie was removed and `false` for when the cookie wasn't. In that case, because of the above conditions, people started expecting impossible results.

A couple historic links for reference:

https://github.com/carhartl/jquery-cookie/issues/382#issuecomment-166923539
https://github.com/carhartl/jquery-cookie/issues/336#issuecomment-72659244
https://github.com/carhartl/jquery-cookie/issues/224

mmcev106: Understood.  I was thinking we could return it for the contexts under which it made sense, but I can see how that would causes too much confusion for the contexts where it doesn't.  Thanks for the response.

FagnerMartinsBrack: Cool, thanks for the PR anyway :)

PR #364, opened 2017-09-21
Do not remove quotes for json

### LINKED ISSUE #363: getJSON does not work on cookie with escaped quotes
Example: The value is `"{\"key1\":\"val1\",\"key2\":\"val2\"}"` including the outermost quotes. Calling `JSON.parse` on this normally would work but these lines remove the quotes at the ends which causes a parse error to occur because the value becomes `{\"key1\":\"val1\",\"key2\":\"val2\"}`.
https://github.com/js-cookie/js-cookie/blob/master/src/js.cookie.js#L112-L114

What is the purpose of those lines? Should they be only for the non-JSON get?

You can test this out with `JSON.parse('"{\\"key1\\":\\"val1\\",\\"key2\\":\\"val2\\"}"')` which works. And `JSON.parse('{\\"key1\\":\\"val1\\",\\"key2\\":\\"val2\\"}')` does not. The double backslash is so that the backslash stays in the resulting string otherwise the `\"` gets evaluated to just `"`.

farazs:
Fixes #363 

I am not sure why quotes are removed to begin with. A cookie might contain double quotes that were intended to be there. I think the lines should be removed altogether but this change fixes my problem.

### DIFF
--- a/test/tests.js
+++ b/test/tests.js
@@ -460,8 +460,8 @@ QUnit.test('Cookies with escaped quotes in json', function (assert) {
 			return value;
 		}
 	}).set('c', '"{ \\"foo\\": \\"bar\\" }"');
-	assert.deepEqual(Cookies.getJSON('c'), '{ "foo": "bar" }', 'returns JSON parsed cookie');
-	assert.deepEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie with enclosing quotes removed');
+	assert.strictEqual(Cookies.getJSON('c'), '{ "foo": "bar" }', 'returns JSON parsed cookie');
+	assert.strictEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie with enclosing quotes removed');
 });
 
 QUnit.module('noConflict', lifecycle);
--- a/test/tests.js
+++ b/test/tests.js
@@ -461,7 +461,7 @@ QUnit.test('Cookies with escaped quotes in json', function (assert) {
 		}
 	}).set('c', '"{ \\"foo\\": \\"bar\\" }"');
 	assert.deepEqual(Cookies.getJSON('c'), '{ "foo": "bar" }', 'returns JSON parsed cookie');
-	assert.deepEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie');
+	assert.deepEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie with enclosing quotes removed');
 });
 
 QUnit.module('noConflict', lifecycle);
--- a/test/tests.js
+++ b/test/tests.js
@@ -451,6 +451,19 @@ QUnit.test('Call to read all cookies with mixed json', function (assert) {
 	assert.deepEqual(Cookies.get(), { c: '{"foo":"bar"}', c2: 'v' }, 'returns unparsed cookies');
 });
 
+QUnit.test('Cookies with escaped quotes in json', function (assert) {
+	Cookies.withConverter({
+		read: function (value) {
+			return value;
+		},
+		write: function (value) {
+			return value;
+		}
+	}).set('c', '"{ \\"foo\\": \\"bar\\" }"');
+	assert.deepEqual(Cookies.getJSON('c'), '{ "foo": "bar" }', 'returns JSON parsed cookie');
+	assert.deepEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie');
+});
+
 QUnit.module('noConflict', lifecycle);
 
 QUnit.test('do not conflict with existent globals', function (assert) {
--- a/test/tests.js
+++ b/test/tests.js
@@ -451,7 +451,7 @@ QUnit.test('Call to read all cookies with mixed json', function (assert) {
 	assert.deepEqual(Cookies.get(), { c: '{"foo":"bar"}', c2: 'v' }, 'returns unparsed cookies');
 });
 
-QUnit.test('Cookies with escaped quotes in json', function (assert) {
+QUnit.test('Cookies with escaped quotes in json using raw converters', function (assert) {
 	Cookies.withConverter({
 		read: function (value) {
 			return value;
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -109,7 +109,7 @@
 				var parts = cookies[i].split('=');
 				var cookie = parts.slice(1).join('=');
 
-				if (cookie.charAt(0) === '"') {
+				if (!this.json && cookie.charAt(0) === '"') {
 					cookie = cookie.slice(1, -1);
 				}
 


test/tests.js
```diff
@@ -451,6 +451,19 @@ QUnit.test('Call to read all cookies with mixed json', function (assert) {
...
 });
 
+QUnit.test('Cookies with escaped quotes in json', function (assert) {
+	Cookies.withConverter({
+		read: function (value) {
+			return value;
+		},
+		write: function (value) {
+			return value;
+		}
```
FagnerMartinsBrack: Why are the converters necessary?

Can you just do:

```javascript
Cookies.set('c', '"{ \\"foo\\": \\"bar\\" }"');
```

farazs: By default the quotes will get escaped which is not the value I want to test. I want the final value to be exactly `"{\"foo\":\"bar\"}"`

If I update the code like you said it fails:
```
>> Message: returns unparsed cookie with enclosing quotes removed
>> Actual: "\"{ \\\"foo\\\": \\\"bar\\\" }\""
>> Expected: "{ \\\"foo\\\": \\\"bar\\\" }"
```

farazs:
What are your thoughts on removing those lines altogether? https://github.com/js-cookie/js-cookie/pull/364/files#diff-b686f7b30e7b0c4f5ed6cd2e87241704R112 to 114

I don't see why they should be in there at all. Quotes around a cookie are perfectly valid, if the caller would like to remove them after the fact they can do so. Right now the caller cannot know whether the original cookie value had quotes around it or not.

test/tests.js
farazs: Also, after removing the converters, the first test case passes but it's not actually testing the case that I am fixing with my change.

FagnerMartinsBrack: Oh, I see, now that makes sense. You want to test the raw converters and the raw `getJSON()`. Thanks for the clarification.

FagnerMartinsBrack:
> [...] What are your thoughts on removing those lines altogether? [...]
> [...] Quotes around a cookie are perfectly valid, if the caller would like to remove them after the fact they can do so [...]

I know where you're coming from and I agree those are great reasons to challenge this. The RFC 6265 accepts quoted cookies.

However, the decision of not supporting quoted cookies was made [5 years ago](https://github.com/js-cookie/js-cookie/commit/ff9dfc634e91a7cbd32a27596efa1fb0199c2d0c) [based on developer request](https://github.com/carhartl/jquery-cookie/pull/57).

In the spirit of backward compatibility and given that we are not planning any major bump anytime soon, do you think it's reasonable to spend the effort to revisit this?

If we decide to revisit this, we'll have to ping everybody involved in this in the past to see their thoughts on it and make sure we're doing the right call.

@carhartl, what are your thoughts on this?

test/tests.js
```diff
@@ -451,6 +451,19 @@ QUnit.test('Call to read all cookies with mixed json', function (assert) {
 	assert.deepEqual(Cookies.get(), { c: '{"foo":"bar"}', c2: 'v' }, 'returns unparsed cookies');
 });
 
+QUnit.test('Cookies with escaped quotes in json', function (assert) {
```
FagnerMartinsBrack: Can we rename this to "Cookies with escaped quotes in json using raw converters"?

test/tests.js
```diff
@@ -451,6 +451,19 @@ QUnit.test('Call to read all cookies with mixed json', function (assert) {
...
+	Cookies.withConverter({
+		read: function (value) {
+			return value;
+		},
+		write: function (value) {
+			return value;
+		}
+	}).set('c', '"{ \\"foo\\": \\"bar\\" }"');
+	assert.deepEqual(Cookies.getJSON('c'), '{ "foo": "bar" }', 'returns JSON parsed cookie');
```
FagnerMartinsBrack: We don't need `deepEqual` here. Deep quality is when we need to compare the structure of Literals such as `[...]` or `{...}`. We don't need that for String which is an immutable data structure. Can we use `assert.strictEqual` instead, so it compares using `===`?

test/tests.js
```diff
@@ -451,6 +451,19 @@ QUnit.test('Call to read all cookies with mixed json', function (assert) {
...
+	assert.deepEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie with enclosing quotes removed');
```
FagnerMartinsBrack: Same here, can we use `strictEqual`?

FagnerMartinsBrack:
@farazs Justa a couple of additional comments in regards some minor things.

carhartl:
> the decision of not supporting quoted cookies 

I don't understand how we are not supporting quoted cookies. Could you give an example of what kind of value is not supported?

If I understand the RFC correctly, double quotes are not allowed as part of the value:

```
cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
                       ; US-ASCII characters excluding CTLs,
                       ; whitespace DQUOTE, comma, semicolon,
                       ; and backslash
```

So is the question is whether to be generally RFC 6265 compliant or not?

Kind of similar discussion here, the reasoning of which I'm leaning towards: https://github.com/golang/go/issues/10195

farazs:
@carhartl Double quotes seem to be permitted around the value. See also this stack overflow post: https://stackoverflow.com/questions/6108207/definite-guide-to-valid-cookie-values

I mentioned in #363 but I'll repeat here:
Example: The value is `"{\"key1\":\"val1\",\"key2\":\"val2\"}"` including the outermost quotes. Calling `JSON.parse` on this normally would work but these lines remove the quotes at the ends which causes a parse error to occur when calling `getJSON` because the value becomes `{\"key1\":\"val1\",\"key2\":\"val2\"}` before the `JSON.parse` is called.

farazs:
Aside from `getJSON` failing, `get` returns the wrong value since it omits the surrounding double quotes.

carhartl:
> Aside from getJSON failing, get returns the wrong value since it omits the surrounding double quotes.

Ok. It looks like lack of RFC 6265 compliance had been incorrectly reported as an issue back then and ever since then my (incorrect) understanding had been that the enclosing double quotes are not to be considered part of the value itself. I need to read the issue again in the go repo, who seem to believe the same. 

If that’s then simply a bug, why don’t you remove it? No need for lengthy discussions.

carhartl:
However - that Stackoverflow page says this:

> The special characters cannot be used at all, with the exception that double quotes may surround the value.

and then again it’s not clear to me whether double quotes are part of the value or just _surround_ the actual value. I‘m sorry, I just don‘t know how to interpret the specification in that regard.

carhartl:
https://github.com/ScottHamper/Cookies/issues/25#issuecomment-43525752

According to that stripping away is giving us the cookie-octet and not the value, which seems to be in line with the specification. So yeah, our implementation then seems to be incorrect, like initially said. So I would just remove that. I don’t even think the real-life implications are that big.. can’t know for sure though.

carhartl:
PS: The original issue was referring to RFC 2068 I finally noticed, so I’m too confused with all of this I guess, I’m sorry. Given that RFC 6265 obsoletes all other specifications we should just support that correctly imo.

farazs:
Also, I think not stripping is better because then you get the raw value. If the user would like to strip enclosing double quotes from all values, they can still do that by wrapping this code. However, the reverse is not true, if I want to add the double quotes back I cannot know which cookies originally had them because that information is lost when I do the `get`.

FagnerMartinsBrack:
@farazs Makes a good point here:

> I think not stripping is better because then you get the raw value. If the user would like to strip enclosing double quotes from all values, they can still do that by wrapping this code. However, the reverse is not true, if I want to add the double quotes back I cannot know which cookies originally had them because that information is lost when I do the get.

We just need to discuss this further.

@farazs if we decide to change this, it will require a major bump (v3.0.0). Can you please create a new issue or PR specifically for this issue and link it here so that we can put it in a new v3 milestone and continue the discussion? Thanks.

I'm tending to avoid coding rules and instead return the whole original value so that the developer decides what to do with it.

In the meantime, I'll merge this fix as long as there's no other objection for this PR only.

FagnerMartinsBrack:
@carhartl @farazs I've created https://github.com/js-cookie/js-cookie/pull/366 to continue this.

Issue #265, opened 2016-09-07
Minfied Version


LordMidi:
Hi!

Would be cool to include a minified version in the project.
Thx for the good work.

FagnerMartinsBrack:
The minified version can be found on the [Releases page](https://github.com/js-cookie/js-cookie/releases) and we have [plans to add to the build but it's not that simple and any help is welcome](https://github.com/js-cookie/js-cookie/issues/57).

PR #394, opened 2017-12-26
Added minified version of the script

MagicMarvMan:
I added a minified version of the script


FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/issues/265.

Issue #396, opened 2018-01-15
Semicolons are accepted in attribute values but they can't be set


macintorsten:
When setting a cookie the API accepts an attribute value that contains a semicolon, and it does not give any error. If the attribute value is coming from an intrusted source it can lead to injection of new attributes than the intended one.

Example: path = '/numbers;domain=.iana.org'
Cookies('nameofcookie', 'cookievalue', {path: '/numbers;domain=.iana.org'})

Reproduce;

> location='https://www.iana.org/protocols'
> $.getScript('https://cdn.jsdelivr.net/npm/js-cookie@2/src/js.cookie.min.js')
> Cookies('nameofcookie', 'cookievalue', {path: '/numbers;domain=.iana.org'})
> console.log(document.cookie)

Output:    [EMPTY]

> location='/numbers'
> console.log(document.cookie)

Output: nameofcookie=cookievalue

I think the path value cannot contain a semicolon but I am not completely sure.

FagnerMartinsBrack:
Ok, cool, here's the long version.

I've tested and confirmed this locally, here are my thoughts:

#### There's no such API

There's no such API as `Cookies(String, String, Object Literal)`. Please, [refer to the docs](https://github.com/js-cookie/js-cookie/tree/a0012f21ba0519fa3c7571cc6f16e15902ec051b#basic-usage) for supported usage and don't use undocumented APIs just because the internals allows you to. [Here's a relevant content](https://medium.com/@fagnerbrack/library-internals-do-not-exist-1ff50a30e3b0) on the subject.

#### The problem is in the application code, not js-cookie

The problem happens when the developer allows an untrusted source to set the cookie attributes. 

The following example can't be exploited if the developer pass an untrusted input in the cookie value due to the default encoding mechanism and the fact RFC 6265 doesn't allow `;` in the cookie value:

```javascript
Cookies.set('nameofcookie', 'cookievalue;path=/path')
```

In this case, the cookie is escaped. The browser creates it as `document.cookie = "nameofcookie=cookievalue%3Bpath=/path"`.

However, the following example can be exploited if the developer uses a raw converter:

```javascript
const RawWriteCookies = Cookies.withConverter({ write: (rawValue) => rawValue });
RawWriteCookies.set('nameofcookie', 'cookievalue;path=/path')
```

In this case, the cookie is created as `document.cookie = "nameofcookie=cookievalue;path=/path"`

[...]

Issue #370, opened 2017-10-21
Remove support for IE 6, 7 and 8


FagnerMartinsBrack:
Some parts of our development environment are starting to break due to many dependencies removing support for IE 8.

Here I'll list all the problems that are happening because of it:

* `qunitjs ` doesn't support IE 6, 7 and 8 and that [prevents js-cookie from upgrading](https://github.com/js-cookie/js-cookie/pull/216#issuecomment-228578080).
* `grunt-contrib-uglify` version 2 [was failing due to `screwIE8` option](https://github.com/js-cookie/js-cookie/pull/242#issuecomment-236415359).

FagnerMartinsBrack:
It looks like Sauce Labs does not support IE7 and IE8 anymore:

<img width="613" alt="screen shot 2017-10-28 at 3 03 33 pm" src="https://user-images.githubusercontent.com/835857/32131089-460175b6-bbf1-11e7-9eec-78d6b4de2b82.png">

FagnerMartinsBrack:
Just to point it out, if we can remove IE 6, 7, 8 support there's a lot of bytes we might be able to trim out of the source :D

tbillington:
Microsoft [stopped supporting IE10 over 2 years ago](https://www.microsoft.com/en-au/windowsforbusiness/end-of-ie-support), and the implementation of cookies is not going to change in older browsers, it seems fair to abandon unsupported browsers when making a new major version.

Users can always stay on version 2 of the library if they need the backwards compatibility.

FagnerMartinsBrack:
> [...] the implementation of cookies is not going to change in older browsers

The cookies implementation in browsers doesn't matter for the purpose of compatibility. It's about which JS features we can use and which ones we can't, that's the library code, not browsers code.

We're definitely considering drop support due to the problems stated above. However, because there's not a big cost to support those browsers in this project, we're just leaving it.

We might as well drop it in the next major release.

ColinLaws:
Well, you're already at ~900 bytes gzipped, so I can't see size as really being a big sticking point. The real question is how big are the improvements you're trying to make, and if support for those browsers is a major feature of your software. I feel like supporting older browsers is more of a testament towards engineering practices than anything, but the minute it starts to hurt when trying to move forward you need to let it go. 

I guess the real question is: Who is js-cookie for? 

[...]

PR #400, opened 2018-02-01
Prevent XSS in the cookie attributes

FagnerMartinsBrack:
Honestly, I'm not happy with this +19 bytes gzipped:

screen shot 2018-02-01 at 10 06 07 pm

X-Ref: https://github.com/js-cookie/js-cookie/issues/396

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -87,6 +87,16 @@
 					if (attributes[attributeName] === true) {
 						continue;
 					}
+
+					// Considers RFC 6265 section 5.2:
+					// ...
+					// 3.  If the remaining unparsed-attributes contains a %x3B (";")
+					//     character:
+					// Consume the characters of the unparsed-attributes up to,
+					// not including, the first %x3B (";") character.
+					// ...
+					attributes[attributeName] = attributes[attributeName].split(';')[0];
+
 					stringifiedAttributes += '=' + attributes[attributeName];
 				}
 				return (document.cookie = key + '=' + value + stringifiedAttributes);
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -95,9 +95,7 @@
 					// Consume the characters of the unparsed-attributes up to,
 					// not including, the first %x3B (";") character.
 					// ...
-					attributes[attributeName] = attributes[attributeName].split(';')[0];
-
-					stringifiedAttributes += '=' + attributes[attributeName];
+					stringifiedAttributes += '=' + attributes[attributeName].split(';')[0];
 				}
 				return (document.cookie = key + '=' + value + stringifiedAttributes);
 			}
--- a/README.md
+++ b/README.md
@@ -203,7 +203,9 @@ Cookies.remove('name', { path: '' });
 
 (From [Internet Explorer Cookie Internals (FAQ)](http://blogs.msdn.com/b/ieinternals/archive/2009/08/20/wininet-ie-cookie-internals-faq.aspx))
 
-This means one cannot set a path using `path: window.location.pathname` in case such pathname contains a filename like so: `/check.html` (or at least, such cookie cannot be read correctly).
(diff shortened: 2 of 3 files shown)


FagnerMartinsBrack:
https://github.com/js-cookie/js-cookie/issues/370#issuecomment-350418712 might help to recover these bytes, though...

src/js.cookie.js
```diff
@@ -87,6 +87,16 @@
...
+					// Considers RFC 6265 section 5.2:
+					// ...
+					// 3.  If the remaining unparsed-attributes contains a %x3B (";")
+					//     character:
+					// Consume the characters of the unparsed-attributes up to,
+					// not including, the first %x3B (";") character.
+					// ...
+					attributes[attributeName] = attributes[attributeName].split(';')[0];
+
 					stringifiedAttributes += '=' + attributes[attributeName];
```
carhartl: We could write
```javascript
stringifiedAttributes += '=' + attributes[attributeName].split(';')[0];
```
and save some more bytes.

carhartl: bildschirmfoto 2018-02-02 um 09 46 58

carhartl: Well, unless `attributes[attributeName]` were needed afterwards for some other processing... but it doesn't look like it.

carhartl:
Looking through the code, I found at least one opportunity to reduce bytes, there might be more:

```javascript
key = encodeURIComponent(String(key));
key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
```
=>
```javascript
key = encodeURIComponent(String(key))
    .replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent)
    .replace(/[\(\)]/g, escape);
```

I'm going to prepare a PR for it.

FagnerMartinsBrack:
> I'm going to prepare a PR for it.

X-Ref: https://github.com/js-cookie/js-cookie/pull/401

src/js.cookie.js
FagnerMartinsBrack: Good idea! Done.

FagnerMartinsBrack:
Bytes after https://github.com/js-cookie/js-cookie/pull/400/commits/99e379b85c659f6937f2e2facedbec1dcc28c344:

screen shot 2018-02-02 at 10 15 35 pm

I guess `+6` and `-6` from https://github.com/js-cookie/js-cookie/pull/401 negates each other :D

carhartl:
Haha, so I'll have to find a way to get rid of 1 more byte...

carhartl:
We have 22 on the plus side now...

qiqingfu:
@carhartl 

hi, I look at the code in the process, encountered a question, but also ask edify ing on the . I can understand this line of code in this function, but I don't understand why you're splitting up.

attributes are an option object that sets cookies, but the value value of this option object contains ';'' .

For example, I write like this.

```js
const attributes = {
  path: '/',
  expires: 7,
}
Cookie.set('name', 'value', attributes)
```
The value values of the attributes are '/' and 7, respectively, and do not contain ';'' the symbol, why split it. this makes me a little confused, I hope the big guy to tell you what you think.

Thank you very much.

FagnerMartinsBrack:
@qiqingfu If you remove those lines, run the tests and debug, you'll find out why they're necessary. IF no test fails, then that means they're not necessary as we cover 100% of the use cases.

qiqingfu:
@FagnerMartinsBrack 

Thank you very much for your answer, I found the reason in the test case.

```js
QUnit.test(
  'sanitization of attributes to prevent XSS from untrusted input',
  function (assert) {
    assert.expect(1)
    assert.strictEqual(
      Cookies.set('c', 'v', {
        path: '/;domain=sub.domain.com',
        domain: 'site.com;remove_this',
        customAttribute: 'value;;remove_this'
      }),
      'c=v; path=/; domain=site.com; customAttribute=value',
      'should not allow semicolon in a cookie attribute'
    )
  }
)
```

Issue #286, opened 2016-12-10
The Build Status Matrix is showing failing tests for a PR created by greenkeeper


FagnerMartinsBrack:
At this moment, the [Build Status Matrix](https://github.com/js-cookie/js-cookie/tree/7fd36e66b60a4e0ac9d09207fb3a348a2969152c#build-status-matrix) seems to show the information for the Sauce Labs of Pull Requests created by greenkeeper ([this one](https://github.com/js-cookie/js-cookie/pull/285#partial-pull-merging), for example). Sometimes, the matrix can show a failing browser in the `master` branch, but in reality what is failing is a PR, not `master`. A developer looking at the README will assume that the library is broken when in reality it isn't.

If we manage to show only the `master` branch Build Status Matrix in the README, then the information will be more precise. It doesn't matter if the same information is shown the README link of all releases, for example, [for this link](https://github.com/js-cookie/js-cookie/tree/v2.1.3#build-status-matrix), [this link](https://github.com/js-cookie/js-cookie/tree/v2.1.0#build-status-matrix) and [this link](https://github.com/js-cookie/js-cookie/tree/v1.5.1#build-status-matrix). Ideally, we should only show the information of the build for that release.

FagnerMartinsBrack:
I've sent a message to Sauce Labs support to see if they can help out to fix this problem.

FagnerMartinsBrack:
They suggesed to submit a feature request on https://saucelabs.ideas.aha.io/

FagnerMartinsBrack:
I have create an idea on Sauce Labs. Let's see what happens.

https://saucelabs.ideas.aha.io/ideas/INTG-I-25

(Anyone arriving here, feel free to go there and upvote it to increase priority)

FagnerMartinsBrack:
We should probably wait for some months. If there's no response from Sauce Labs, add a note in the README to document the issue and close this.

FagnerMartinsBrack:
It's been more than a year. Let's give up.

@carhartl is trying to migrate to [BrowserStack](https://github.com/js-cookie/js-cookie/pull/409). Maybe this issue will be fixed there?

FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/commit/8b311b9dab5751fab3b09b58d215262e908e92e1. I have changed the README to include a note that the Build Status Matrix includes active Pull Requests.

PR #401, opened 2018-02-02
Reduce bytes

carhartl:
The goal is to reduce bytes without compromising on readability.

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -102,11 +102,12 @@
 				result = {};
 			}
 
+			var decode = function (s) {
+				return s.replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent);
+			};
 			// To prevent the for loop in the first place assign an empty array
-			// in case there are no cookies at all. Also prevents odd result when
-			// calling "get()"
+			// in case there are no cookies at all.
 			var cookies = document.cookie ? document.cookie.split('; ') : [];
-			var rdecode = /(%[0-9A-Z]{2})+/g;
 			var i = 0;
 
 			for (; i < cookies.length; i++) {
@@ -118,10 +119,9 @@
 				}
 
 				try {
-					var name = parts[0].replace(rdecode, decodeURIComponent);
-					cookie = converter.read ?
-						converter.read(cookie, name) : converter(cookie, name) ||
-						cookie.replace(rdecode, decodeURIComponent);
+					var name = decode(parts[0]);
+					cookie = (converter.read || converter)(cookie, name) ||
+						decode(cookie);
 
 					if (this.json) {
 						try {
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -143,7 +143,7 @@
 		api.getJSON = function () {
 			return api.apply({
 				json: true
-			}, [].slice.call(arguments));
+			}, arguments);
 		};
 		api.remove = function (key, attributes) {
 			api(key, '', extend(attributes, {
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -6,7 +6,7 @@
  * Released under the MIT license
  */
 ;(function (factory) {
-	var registeredInModuleLoader = false;
+	var registeredInModuleLoader;
 	if (typeof define === 'function' && define.amd) {
 		define(factory);
 		registeredInModuleLoader = true;
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -38,7 +38,6 @@
 
 	function init (converter) {
 		function api (key, value, attributes) {
-			var result;
 			if (typeof document === 'undefined') {
 				return;
 			}
@@ -58,7 +57,7 @@
 				attributes.expires = attributes.expires ? attributes.expires.toUTCString() : '';
 
 				try {
-					result = JSON.stringify(value);
+					var result = JSON.stringify(value);
 					if (/^[\{\[]/.test(result)) {
 						value = result;
 					}
@@ -98,10 +97,7 @@
 
 			// Read
 
-			if (!key) {
-				result = {};
-			}
-
+			var jar = {};
 			var decode = function (s) {
 				return s.replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent);
 			};
@@ -129,18 +125,15 @@
 						} catch (e) {}
 					}
 
-					if (key === name) {
-						result = cookie;
-						break;
-					}
+					jar[name] = cookie;
 
-					if (!key) {
-						result[name] = cookie;
+					if (key === name) {
+						break;
 					}
 				} catch (e) {}
 			}
 
-			return result;
+			return key ? jar[key] : jar;
 		}
 
 		api.set = api;
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -73,9 +73,9 @@
 					value = converter.write(value, key);
 				}
 
-				key = encodeURIComponent(String(key));
-				key = key.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent);
-				key = key.replace(/[\(\)]/g, escape);
+				key = encodeURIComponent(String(key))
+					.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent)
+					.replace(/[\(\)]/g, escape);
 
 				var stringifiedAttributes = '';
 				for (var attributeName in attributes) {
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -66,12 +66,10 @@
 					}
 				} catch (e) {}
 
-				if (!converter.write) {
-					value = encodeURIComponent(String(value))
+				value = converter.write ?
+					converter.write(value, key) :
+					encodeURIComponent(String(value))
 						.replace(/%(23|24|26|2B|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g, decodeURIComponent);
-				} else {
-					value = converter.write(value, key);
-				}
 
 				key = encodeURIComponent(String(key))
 					.replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent)
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -51,9 +51,7 @@
 				}, api.defaults, attributes);
 
 				if (typeof attributes.expires === 'number') {
-					var expires = new Date();
-					expires.setMilliseconds(expires.getMilliseconds() + attributes.expires * 864e+5);
-					attributes.expires = expires;
+					attributes.expires = new Date(new Date() * 1 + attributes.expires * 864e+5);
 				}
 
 				// We're using "expires" because "max-age" is not supported by IE


carhartl:
bildschirmfoto 2018-02-02 um 10 19 34

carhartl:
Ah, the latest commit doesn't seem to help with the gzipped sized... so I can just as well remove it again from the PR, though I like how `key` + `value` assignments stand on the same line now.

bildschirmfoto 2018-02-02 um 10 47 06

carhartl:
No, if I interprete it correctly, compared to the latest minified release, 2.2.0, we'll get a decrease of 6 bytes when doing a new release with these changes?

FagnerMartinsBrack:
> ... though I like how key + value assignments stand on the same line now

I have just one point that's probably a matter of style, also probably be subjective and I have no evidence to support it:

I've seen people having trouble to understand a codebase with lots of ternary statements. An explicit `if` is easier to reason about. I would say that, because it's subjective, we can sacrifice explicit `if`s in favor of ternary, but when it doesn't make any difference is it worth keeping the explicit `if`?

Not strong about it, though. It achieved a good result in reducing 5 bytes :)

carhartl:
We're using said operator in other places as well, those weren't replaced with explicit if statements either. Thus, imo, the code is even more consistent now. I heard the argument about ternary vs if before and won't deny it, but readability in this case much depends on how huge the operands get and/or whether you start nesting ternary operators. Consistency is equally important  to me.

carhartl:
I'm not merging yet, because I wanted to take another look, maybe I find another place to take away some bytes.

FagnerMartinsBrack:
> ... but readability in this case much depends on how huge the operands get and/or whether you start nesting ternary operators

Agreed

carhartl:
We could do
```javascript
if (typeof attributes.expires === 'number') {
    attributes.expires = new Date(new Date().getTime() + attributes.expires * 864e+5);
}
```
and this would result in reduction of another 17 bytes. I'm not too happy about creating two Date instances, but does it really matter? I would have used `Date.now()` instead of `new Date().getTime()` but that is only supported in IE 11...

carhartl:
`new Date(new Date() * 1 + attributes.expires * 864e+5)`, ha

FagnerMartinsBrack:
I don't think there's any problem with more `Date` instances. There are limits for cookies so one is not supposed to be calling `Cookies.get()` many times.

carhartl:
I guess that would be ok because that line in particular was already obfuscated by that `attributes.expires * 864e+5`.... :)

carhartl:
bildschirmfoto 2018-02-02 um 12 57 47

carhartl:
I have to agree though `new Date(new Date() * 1 + attributes.expires * 864e+5)` interferes with:

> The goal is to reduce bytes without compromising on readability.

🙈

FagnerMartinsBrack:
Honestly, as long as it has tests and doesn't break other use cases, I'm fine with it. This codebase is heavily trading off legibility for less bytes already so I don't think it makes much difference.

carhartl:
Interesting, I was experimenting with making the `extend()` function smaller:
```javascript
function extend () {
	var i = 0;
	var result = {};
	while (arguments[i]) {
		for (var key in arguments[i]) {
			result[key] = arguments[i][key];
		}
		i++;
	}
	return result;
}
```
With that I'm getting a test failure
```
>> remove - deletion
>> Message: should delete the cookie
>> Actual: "c="
>> Expected: ""
```
and I can't figure out why! That function works fine when experimenting in the (node) console, e.g. it produces the same result as the former one... what am I overlooking?

carhartl:
Ah, it was because  `attributes` was `undefined` in the test case.

carhartl:
..and then the changes would turn into a `-21` 👎 

This is fun. :)

carhartl:
We could change collecting the result(s) to this, would increase savings to `27`:
```patch
diff --git a/src/js.cookie.js b/src/js.cookie.js
index 4618358..5a90898 100644
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -38,7 +38,6 @@
 
 	function init (converter) {
 		function api (key, value, attributes) {
-			var result;
 			if (typeof document === 'undefined') {
 				return;
 			}
@@ -58,7 +57,7 @@
 				attributes.expires = attributes.expires ? attributes.expires.toUTCString() : '';
 
 				try {
-					result = JSON.stringify(value);
+					var result = JSON.stringify(value);
 					if (/^[\{\[]/.test(result)) {
 						value = result;
 					}
@@ -90,16 +89,13 @@
 
 			// Read
 
-			if (!key) {
-				result = {};
-			}
-
 			// To prevent the for loop in the first place assign an empty array
 			// in case there are no cookies at all. Also prevents odd result when
 			// calling "get()"
 			var cookies = document.cookie ? document.cookie.split('; ') : [];
 			var rdecode = /(%[0-9A-Z]{2})+/g;
 			var i = 0;
+			var jar = {};
 
 			for (; i < cookies.length; i++) {
 				var parts = cookies[i].split('=');
@@ -121,18 +117,15 @@
 						} catch (e) {}
 					}
 
+					jar[name] = cookie;
+
 					if (key === name) {
-						result = cookie;
 						break;
 					}
-
-					if (!key) {
-						result[name] = cookie;
-					}
 				} catch (e) {}
 			}
 
-			return result;
+			return key ? jar[key] : jar;
 		}
 
 		api.set = api;
```

FagnerMartinsBrack:
Sounds great! You're having a lot of fun ain't you? :D

carhartl:
Another thing I noticed... @FagnerMartinsBrack 

We have
```javascript
attributes = extend({
    path: '/'
}, api.defaults, attributes);
```
but why is `path: '/'` not part of `defaults` right away (de facto it is a default)?

carhartl:
Ah, the test for `'API for changing defaults` told me...

carhartl:
`-25`

I'm reluctant to remove 
```javascript
if (key === name) {
    break;
}
```
from the loop, which would give us `-35`, but would have aforementioned performance implications.

carhartl:
`-30` was my goal, we have `-33` now.

src/js.cookie.js
```diff
@@ -146,16 +135,16 @@
 		api.getJSON = function () {
 			return api.apply({
 				json: true
-			}, [].slice.call(arguments));
+			}, arguments);
```
carhartl: Could this be problematic in older browsers?

FagnerMartinsBrack: We don't run the tests on IE 6 and 7 anybody [because of reasons](https://github.com/js-cookie/js-cookie/issues/370#issuecomment-340137482). IE 8 seems to be working, though.

Need to manually test in those browsers I guess because the version we're working on still supports it in the docs.

carhartl:
I‘m pretty sure my old Parallels installation which allowed me to run IE 6 + 7 on Windows does no longer work. Are there still official images provided by MS allowing to test these?

src/js.cookie.js
carhartl: Fwiw, the only reason I could think of why we had the arguments to array trick in there was because of JavaScript engines in these old browsers. Unfortunately I can’t remember whether IE had a problem with that or not.

FagnerMartinsBrack:
> Are there still official images provided by MS allowing to test these?

I don't know. I've only relied on Sauce Labs and landed on master things that I was considerably confident it wouldn't break in IE6/7...

FagnerMartinsBrack:
By the way, I'm reviewing each commit, running the tests locally, checking byte size, etc. and I'll cherry-pick them to master directly.

Probably change the message to add the gain in bytes for each commit...

FagnerMartinsBrack:
Actually... given the commits are very well organized we might as well merge into master instead of rebasing... Maybe update the commit messages to tell how many bytes each of them removed?

carhartl:
https://www.browserstack.com/ says it’s free for open source. If it supports IE 6 + 7 we could switch, though I’m not sure it’s worth it. I’d rather remove support for those  anytime soon.

FagnerMartinsBrack:
> If it supports IE 6 + 7 we could switch

It's a lot of effort to switch :D. Can you use the runnable version from them and run the tests on IE 6? That should be enough I guess...

FagnerMartinsBrack:
Once you're comfortable with this feel free to do a `git merge`. In the meantime, I''l land https://github.com/js-cookie/js-cookie/pull/400. There might be conflicts, though.

carhartl:
> feel free to do a git merge

Not sure I understand you. Are you asking to merge these changes manually instead of going ahead an merging it right here?

carhartl:
Regarding Browserstack, it's not possible testing IE 6 + 7 without buying a plan:
bildschirmfoto 2018-02-11 um 10 04 03

My Parallels does in fact no longer work, the (old) version I have is not working in High Sierra. I don't have an idea how to test that. I do not even have any licensed Windows XP version any longer.

FagnerMartinsBrack:
> Not sure I understand you. Are you asking to merge these changes manually instead of going ahead and merging them right here?

No no. Feel free to merge from here. The commits from this PR [are atomic](https://medium.com/@fagnerbrack/one-commit-one-change-3d10b10cebbf) so it should be fine to have them on `master`.

As a sidenote:

What I have been done so far is to close the PR from a commit on master with the message `Closes gh-xx` where `xx` is the target PR number ([see here](https://github.com/js-cookie/js-cookie/commit/31b3366b0384ddc41558a4bf84ce282e65c27588) for an example). Things are much easier if you [fetch all the PRs from upstream](https://medium.com/@fagnerbrack/one-commit-one-change-3d10b10cebbf), then it's just a matter of doing `git cherry-pick <hash>` on `master` and push it.

The reason I've been doing this is to be able to standardize the commit messages on `master`. Github has changed several times how the merge/rebase message is created.

However, it should be fine to merge from the Github UI with or without Squash. Up to you.

carhartl:
> Regarding Browserstack, it's not possible testing IE 6 + 7 without buying a plan

I've signed up and contacted them, asking for an open source plan. It's possible to use it with Travis. In case we will get access, I would prepare everything to switch from Saucelabs.

carhartl:
Tried converting my Paralles images to VirtualBox but all I got is a bluescreen.

carhartl:
Running out of ideas, I believe I'm going to wait a little more and see whether I get to test on BrowserStack... in the meantime I'm dreaming of removing support for Internet Explorer 6  + 7.

FagnerMartinsBrack:
> I'm dreaming of removing support for Internet Explorer 6 + 7

We can do it, it will just require a major bump to 3.0.0, that's it. What if you leave the `[].slice` change to [milestone `3.0.0`](https://github.com/js-cookie/js-cookie/milestone/11) and then commit everything else and [release the version `2.2.1`](https://github.com/js-cookie/js-cookie/milestone/10?closed=1)?

carhartl:
Oh, ok, it seems you can just download Windows 7 from https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/, and in Windows 7 itself there's a possibility to run Windows XP.

FagnerMartinsBrack:
> If you think it's worth pursuing further, let me know, @FagnerMartinsBrack!

I think it is if you're up to it :D.

Do they have a fix [for this](https://github.com/js-cookie/js-cookie/issues/286)? Can they support IE6+?

carhartl:
> Do they have a fix for this? 
I hope, can‘t tell yet. I also remember we had problems running the full test suite in Saucelabs for PRs from outside the original repo, is that still the case?


> Can they support IE6+?
Yes, they do! (But not in the trial, only for paid plans and, as far as I understood open source plans.)

carhartl:
>I also remember we had problems running the full test suite in Saucelabs for PRs from outside the original repo

Unless we choose to keep the access key in plain text, which we don’t, nope!

https://docs.travis-ci.com/user/browserstack/

carhartl:
At the moment the test suite doesn't even run in IE6, because of:

https://github.com/js-cookie/js-cookie/blob/f9f3f0eb93528214bdc4fda062717eba1861bb45/test/utils.js#L38

carhartl:
But once that's fixed, I conclude this branch works fine in IE 6 with all the changes.

carhartl:
> But once that's fixed

=> 6e42d24f8bf0080949f37db8191516967e6d72b4

carhartl:
Browserstack is amazing. I'm working on moving our testing in CI to it next.

PR #406, opened 2018-02-09
Add package-lock.json file to git ignored files

carhartl:
I think the `package-lock.json` file is better to be ignored? It was sitting in my repo as added when it seemingly shouldn't.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
 build
+package-lock.json
 .sizecache.json
 *.log*


FagnerMartinsBrack:
Since you're touching nodeJS related stuff, can you change the [`engines` property](https://docs.npmjs.com/files/package.json#engines) to tell which version is generating the `package-lock.json`? Otherwise one might install new packages without updating the lock, or updating it incorrectly, in case the node version is different. The "engines" property communicates which version it's supposed to be used.

Also, we might want to update the node version [that is running on Travis](https://github.com/js-cookie/js-cookie/blob/73476a4ecfa5ac3621df501af22e613725e88976/.travis.yml).

carhartl:
Ah, now I get it. I was using a newer node version when running `npm install`, that's where the seemingly unignored file came from.

FagnerMartinsBrack:
It's totally fine to source it, it helps to lock development dependencies into a specific version. If it changes (because inner dependencies uses a prefix like `^` or `~`) then it will mark the file as changes when you `npm install`.

Just have to document the node version somewhere to make sure we generate the same `package-lock.json`.

Issue #226, opened 2016-06-25
FAQ: Cookie obfuscates another one with the same name


FagnerMartinsBrack:
Previous links I could find with issues related in creating 2 cookies with the same name:
- carhartl/jquery-cookie#21
- carhartl/jquery-cookie#172
- carhartl/jquery-cookie#343
- carhartl/jquery-cookie#249
- carhartl/jquery-cookie#178
- carhartl/jquery-cookie#130
- js-cookie/js-cookie#52
- js-cookie/js-cookie#81
- js-cookie/js-cookie#98
- js-cookie/js-cookie#112
- js-cookie/js-cookie#152

We need to test this behavior in the major browsers (Safari, Chrome, Edge, Internet Explorer and Firefox) and document it.

I don't understand exactly how this happens and in which circumstances.

FagnerMartinsBrack:
In order to document this properly, it's necessary to understand why and how a cookie can obfuscate each other.

That means you need to create a test server locally, set many cookies with the same name and observe the behavior.

FagnerMartinsBrack:
According to https://github.com/js-cookie/js-cookie/issues/388#issuecomment-363125036, it looks like "the cookie with the domain that comes first alphabetically is used" when there are duplicates.

FagnerMartinsBrack:
Update: I believe this is still important, mostly to understand how this behavior works across browsers and [hide the leaky abstraction](https://medium.com/@fagnerbrack/repairing-the-leaky-abstraction-e726baab91b5) through [documentation](https://hackernoon.com/how-to-learn-with-open-source-api-fundamentals-1f886383b8e1).

FagnerMartinsBrack:
https://github.com/js-cookie/js-cookie/wiki/Frequently-Asked-Questions#can-i-set-two-cookies-with-the-same-name

PR #418, opened 2018-03-21
Fix getJSON() accidentally writing cookie with too many arguments

### LINKED ISSUE #417: getJSON called with many arguments doesn't behave as expected
Following up on #416 

@carhartl
> getJSON(), in contrast to get(), we‘re passing on all given arguments. So what‘s actually happening when you called it with two arguments is each time a cookie is written (due to the api’s underlying implementation). You won’t notice the difference if the values remain the same, but that seems to be a bug. cc @FagnerMartinsBrack 

@carhartl 
> We have:

```
api.getJSON = function () {
    return api.apply({
        json: true
	}, arguments);
};
```
> and I believe we should have:

```
api.getJSON = function (key) {
    return api.call({ json: true }, key);
};
```

@carhartl 

> I believe it’s worse..

```
cookies.getJSON(k, {domain: 'a'})
```
> would result in a cookie k written with a value of json stringified {domain: 'a'}. Cannot test this right now.

carhartl:
Addresses #417 (for context see there).

- [x] Fix implementation
- [x] Brushup documentation

~Note: I needed to change a test (see e22693488a6c31e9f54d5f4734b92968ace8ea2e), unrelated to the PR, because it was failing due to upcoming DST change, and the expires date was lying after that change, resulting in a date string mismatch. I reduced the likelihood of this happening again by reducing the amount of days expires will be ahead of "now", but there should be a way how to make this proof once and for all.~[-> commit went out of PR to master]

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,16 @@ Read all visible cookies:
 Cookies.get(); // => { name: 'value' }
 ```
 
+*Note: It is not possible to read a particular cookie by passing one of the cookie attributes (which may or may not
+have been used when writing the cookie in question):*
+
+```javascript
+Cookies.get('foo', domain: { 'sub.example.com' }); // `domain` won't have any effect...!
+```
+
+The cookie with the name `foo` will only be available on `.get()` if it's visible from where the
+code is called; the domain and/or path attribute will not have an effect when reading.
+
 Delete cookie:
 
 ```javascript
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -140,10 +140,10 @@
 		api.get = function (key) {
 			return api.call(api, key);
 		};
-		api.getJSON = function () {
-			return api.apply({
+		api.getJSON = function (key) {
+			return api.call({
 				json: true
-			}, arguments);
+			}, key);
 		};
 		api.remove = function (key, attributes) {
 			api(key, '', extend(attributes, {
--- a/test/tests.js
+++ b/test/tests.js
@@ -474,6 +474,11 @@ QUnit.test('Cookies with escaped quotes in json using raw converters', function 
 	assert.strictEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie with enclosing quotes removed');
 });
 
+QUnit.test('Prevent accidentally writing cookie when passing unexpected argument', function (assert) {
+	Cookies.getJSON('c', { foo: 'bar' });
+	assert.strictEqual(Cookies.get('c'), undefined, 'should not write any cookie');
+});
+
 QUnit.module('noConflict', lifecycle);
 
 QUnit.test('do not conflict with existent globals', function (assert) {


test/tests.js
```diff
@@ -180,10 +180,10 @@ QUnit.test('undefined', function (assert) {
 
 QUnit.test('expires option as days from now', function (assert) {
 	assert.expect(1);
-	var sevenDaysFromNow = new Date();
-	sevenDaysFromNow.setDate(sevenDaysFromNow.getDate() + 21);
-	var expected = 'expires=' + sevenDaysFromNow.toUTCString();
-	var actual = Cookies.set('c', 'v', { expires: 21 });
+	var expires = new Date();
```
carhartl: "Seven days from now" had become twenty-one days from now in the meantime, so I changed the naming.

carhartl:
@FagnerMartinsBrack Let me know whether we should first fix the expires related test in a separate commit/PR (-> e22693488a6c31e9f54d5f4734b92968ace8ea2e), and remove it from this PR.

carhartl:
Brushing up documentation as part of this PR?

FagnerMartinsBrack:
> @FagnerMartinsBrack Let me know whether we should first fix the expires related test in a separate commit/PR (-> e226934), and remove it from this PR.

I think you can commit on `master` the fix that this PR is dependent upon. I'm having the same test failure 😆

src/js.cookie.js
```diff
@@ -140,10 +140,10 @@
 		api.get = function (key) {
 			return api.call(api, key);
 		};
-		api.getJSON = function () {
-			return api.apply({
+		api.getJSON = function (key) {
+			return api.call({
```
FagnerMartinsBrack: Any byte size impact or none?

FagnerMartinsBrack:
<sup>I wonder how many contributors are not contributing due to this test failure 😞😞😞 </sup>

carhartl:
> I wonder how many contributors are not contributing due to this test failure in the local dev environment

😢 I'm going to look for a proper solution next.

src/js.cookie.js
carhartl: I haven't had coffee, but I think it's an improvement even:

bildschirmfoto 2018-03-21 um 07 41 04

carhartl:
Regarding the expires test -> #419

carhartl:
So there were 42 days in a year where this test was failing... 😭 

(At least we've reduced it to 2 now.)

src/js.cookie.js
FagnerMartinsBrack: o.O wow

FagnerMartinsBrack:
We should probably close off v2.2.1 with [these issues](https://github.com/js-cookie/js-cookie/milestone/10).

test/tests.js
```diff
@@ -474,6 +474,11 @@ QUnit.test('Cookies with escaped quotes in json using raw converters', function
 	assert.strictEqual(Cookies.get('c'), '{ \\"foo\\": \\"bar\\" }', 'returns unparsed cookie with enclosing quotes removed');
 });
 
+QUnit.test('Prevent accidentally writing cookie when passing unexpected argument', function (assert) {
+	Cookies.getJSON('c', { foo: 'bar' });
+	assert.strictEqual(document.cookie, '', 'should not write any cookie');
```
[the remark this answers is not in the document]
carhartl: I've fixed this...

carhartl:
@FagnerMartinsBrack Added the documenation part... do you think it's sufficient?

README.md
```diff
@@ -95,6 +95,18 @@ Read all visible cookies:
...
+have been used when writing the cookie in question):*
+
+```javascript
+Cookies.get('foo', domain: 'sub.example.com'); // `domain` won't have any effect...!
+```
+
+That is, if a cookie with the name `foo` is visible from where `get()` was called,
+it's the one that is being read, no matter the domain or path. This also means that no guarantee
+can be made regarding which cookie is being retrieved in case one cookie is shadowing another,
+similarly named one with a differently scoped `domain` and/or `path` attribute.
```
FagnerMartinsBrack: I'm not sure of the shadowing case should be documented here? Is it related to the issue we're trying to solve here? Have you tested it?

We have [an issue](https://github.com/js-cookie/js-cookie/issues/226) to document the shadowing cookie problem.

README.md
```diff
@@ -95,6 +95,18 @@ Read all visible cookies:
...
 ```
 
+*Note: It is not possible to read a particular cookie by passing one of the cookie attributes (which may or may not
+have been used when writing the cookie in question):*
+
+```javascript
+Cookies.get('foo', domain: 'sub.example.com'); // `domain` won't have any effect...!
+```
+
+That is, if a cookie with the name `foo` is visible from where `get()` was called,
```
FagnerMartinsBrack: Could we restructure like:

```
The cookie with the name `foo` will only be available on `.get()` if it's visible from where the code is called.
```

README.md
```diff
@@ -95,6 +95,18 @@ Read all visible cookies:
 Cookies.get(); // => { name: 'value' }
 ```
 
+*Note: It is not possible to read a particular cookie by passing one of the cookie attributes (which may or may not
+have been used when writing the cookie in question):*
+
+```javascript
+Cookies.get('foo', domain: 'sub.example.com'); // `domain` won't have any effect...!
```
FagnerMartinsBrack: Missing braces: `Cookies.get('foo', { domain: 'sub.example.com' });`

README.md
```diff
@@ -95,6 +95,18 @@ Read all visible cookies:
 Cookies.get(); // => { name: 'value' }
 ```
 
+*Note: It is not possible to read a particular cookie by passing one of the cookie attributes (which may or may not
```
FagnerMartinsBrack: "It is not possible to `.get()` a particular cookie..."?

FagnerMartinsBrack:
@carhartl I've made some comments

README.md
carhartl: In plain text I don't see what's wrong with using "read"... that's how the operation is usually referred to imo.

FagnerMartinsBrack: That's fine, just wondering if it would make more sense. LGTM

Issue #424, opened 2018-04-01
LICENSE - Fagner Brack not listed in LICENSE file but is in js-cookie.js under "src"


PrestonL:
Fagner Brack not listed in LICENSE file but is in js-cookie.js under "src", copyright dates don't match.

FagnerMartinsBrack:
I'm not a lawyer so I don't know why that matters. @PrestonL, do you want to create a PR updating those files?

PrestonL:
Done with PR 425.  I'll do a little digging into CLA / CAA so that persons doing PRs will have transferred Copyright for MIT if you're interested in shoring that up; right now there may be some issues with that if there are other contributions pushed into the project.

FagnerMartinsBrack:
> right now there may be some issues with that if there are other contributions pushed into the project.
Which issues could that be?

PrestonL:
I am not a lawyer, but the way I understand it is that if someone contributes code to a project, they still retain copyright rights to the derivative work; changes to MIT code aren't necessarily licensed under the MIT License. A contributor license agreement generally assigns copyright to the project maintainer or an open source foundation so that said maintainer can then release the derivative work under the MIT License.  Right now in open source there is a mess of CLAs and CAAs out there.  I'm going to write one for myself and my own F/OSS projects, and can post it if you'd like to implement it yourself; I looked at the FSF and other places, and a good CLA/CAA guide doesn't really seem to exist.

FagnerMartinsBrack:
Do those rules change per country? I guess Github is in US so we're bound to their laws?

> I'm going to write one for myself and my own F/OSS projects, and can post it if you'd like to implement it yourself

If you want to create good guidelines or tools for this we would definitely implement here.

Honestly, I'm not too worried about this. Anybody contributing to Open Source should be aware their work is going to be of the same license as the project. If anybody tries to sue or anything, it's just a matter of reverting the commit. Twbs had a similar problem in the past and they looked at the commits for people who didn't want to change the license and just reverted it.

Unless I'm missing something.

PrestonL:
It gets complicated, copyright is generally similar between counties because of treaties, and because of what you said about the intent of people contributing code, it's mostly a non-issue.

[...]

PR #425, opened 2018-04-02
Update LICENSE

PrestonL:
Updated LICENSE file per conversation with project maintainer

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2014 Klaus Hartl
+Copyright 2006, 2014, 2015 Klaus Hartl & Fagner Brack
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2014 Klaus Hartl
+Copyright 2006, 2014, 2015 Klaus Hartl & Fagner Brack
```
FagnerMartinsBrack: Why do we need to specify the dates like `2006`, `2014` and `2015`? Can't we just put 2018?

FagnerMartinsBrack:
X-Ref: https://github.com/js-cookie/js-cookie/issues/424.

@PrestonL, given our conversation on https://github.com/js-cookie/js-cookie/issues/424, is there any other action to do in this regard?

PrestonL:
I would recommend:

Copyright 2006-2018 Klaus Hartl, Fagner Brack, GitHub Contributors

or even with just the current year:

Copyright 2018 Klaus Hartl, Fagner Brack, GitHub Contributors

I believe the GitHub Terms of Service "Contributions Under Repository License" clause sufficiently covers everything and agree that no CLA is needed as discussed in #424 

The Terms of Service don't necessarily transfer copyright ownership of contributions to the project or maintainer, so attributing all the contributors as a whole does seem it would be an appropriate practice as OpenStreetMap does (they always say Copyright OpenStreetMap Contributors); has a side effect of gives good vibes by crediting the contributors as well.

Glad we were all able to dig into and learn something on that.

Thanks for all your work!

FagnerMartinsBrack:
@PrestonL Do you wanna make the change then? After that change, it LGTM. Let's keep it for a while to see if there's any additional feedback.

Can you post a permalink to OpenStreetMap License or other examples for future context?

PrestonL:
Sure, I created a new pull request and referenced this one in it.  Here is the example license and attribution format used by Open Street Map for reference:

https://www.openstreetmap.org/copyright

FagnerMartinsBrack:
You don't need to create a new Pull Request, just push changes to the branch `patch-1` of your fork.

PR #429, opened 2018-04-12
Make expires test agnostic to DST change

### LINKED ISSUE #419: Make expires related test fail-safe
> I was seeing an error due to a used expires date coming after the
upcoming DST change. Reducing the days we move forward should at least
reduce the likelihood that the test fails.

See 8b2a8801cc53fe9e150ef41c5c79d1193c79a385

We need to make this test fail-safe no matter the timezone or DST.

carhartl:
The test setup itself was assuming a different implementation. While `setDate()` will respect values causing the new date to overlap a DST change, e.g. adapt the hour accordingly, the underlying implementation does not. Therefore we need to set up the expected date to assert for differently, in line with the implementation that is.

Closes #419

### DIFF
--- a/test/tests.js
+++ b/test/tests.js
@@ -180,10 +180,10 @@ QUnit.test('undefined', function (assert) {
 
 QUnit.test('expires option as days from now', function (assert) {
 	assert.expect(1);
-	var expires = new Date();
-	expires.setDate(expires.getDate() + 1);
+	var days = 200;
+	var expires = new Date(new Date().valueOf() + days * 24 * 60 * 60 * 1000);
 	var expected = 'expires=' + expires.toUTCString();
-	var actual = Cookies.set('c', 'v', { expires: 1 });
+	var actual = Cookies.set('c', 'v', { expires: days });
 	assert.ok(actual.indexOf(expected) !== -1, quoted(actual) + ' includes ' + quoted(expected));
 });
 


carhartl:
Also related: 41b1bad52a029e75f68676024a559dd0142fcb67

carhartl:
And the last related change: 4cd7df2a3a7e8746dfcd44f4377759e2d0a15a31

It looks like we went a full circle.. we broke accounting for DST in ecb597b65e4c477baa2b30a2a5a67fdaee9870ea, we fixed it in 41b1bad52a029e75f68676024a559dd0142fcb67, we broke it again in 4cd7df2a3a7e8746dfcd44f4377759e2d0a15a31 

Ugh!

carhartl:
Having realized that, instead of this PR, I think we should revert the implementation for the "expires as days" option to the one from 41b1bad52a029e75f68676024a559dd0142fcb67, while having a test that  uses a value that would always overlap a DST change to ensure we're not running another circle as soon we've forgotten all of this again :)

Your thoughts, @FagnerMartinsBrack ?

FagnerMartinsBrack:
> Having realized that, instead of this PR, I think we should revert the implementation for the "expires as days" option to the one from 41b1bad, while having a test that uses a value for days that would always overlap a DST change (like in this PR with 200 days) to ensure we're not running another circle as soon we've forgotten all of this again :)

LGTM as long as we have a consistently failing test that doesn't rely on JavaScript `Date` object.

test/tests.js
```diff
@@ -180,10 +180,10 @@ QUnit.test('undefined', function (assert) {
 
 QUnit.test('expires option as days from now', function (assert) {
 	assert.expect(1);
-	var expires = new Date();
-	expires.setDate(expires.getDate() + 1);
+	var days = 200;
+	var expires = new Date(new Date() * 1 + days * 864e+5);
```
FagnerMartinsBrack: Is this dependent on the Date we're in? If so, can we create a new `Date` instance on a specific date? If it's too hard to do in JS, I'm not against introducing a third party devDependency such as [moment](https://www.npmjs.com/package/moment)

carhartl: `new Date()` gives us a timestamp representing milliseconds since January 1970 in UTC, which should be fine for the test and we won't need moment. 

If I revert the implementation to actually account for a DST change, I will need to adapt setting up the `expires` variable here to make use of code that would also account for DST (for instance `Date.prototype.setMilliSeconds()`, otherwise we will get 1 hour mismatches again occasionally. That said, what's important is that the test matches the implementation.

carhartl: I am wondering though - neither options seems more right or wrong to me. If we were, today at 16:00:00, writing a cookie with expires of 200 days..:

With accounting for DST, we'd get an expiration with the hour adapted 
`Mon, 29 Oct 2018 15:00:00 GMT`

Without accounting for DST, we'd get an expiration with the hour unchanged:
`Mon, 29 Oct 2018 16:00:00 GMT`

The latter seems even more expected to me, looking at it now.

carhartl: Given the fix for https://github.com/carhartl/jquery-cookie/issues/331 was more of an earlier attempt to fix the test (nobody asked that the hour when using days for expires would need to account for DST, and no behavior was specified either), and given the current implementation saved us a few bytes, I'm now rather going to suggest to keep the implementation as-is, and merge this PR, which fixes the test and makes it agnostic to the `Date` we're in (e.g. the test is being run at)...

carhartl:
> Why closing? Is this not valid anymore? 

It would no longer have been valid with https://github.com/js-cookie/js-cookie/pull/429#issuecomment-380809305, but I changed my mind: https://github.com/js-cookie/js-cookie/pull/429#discussion_r181096960

test/tests.js
FagnerMartinsBrack: I still find this test very hard to comprehend the intentions behind it. I wonder if it makes sense to stub the variants we have (such as JavaScript `Date`) and return a fixed result if it's called a certain way?

```javascript
test('...', () => {
  window.Date = mockedDate;
  // ...
  // restore stub
})
```
Another option is to have our own "Expires" class that allows the expires to be defined like:

```javascript
Cookies.set('foo', { expires: Cookies.Expires.minutes(7) });
Cookies.set('foo', { expires: Cookies.Expires.days(24) });
```

This way we can stub the expires by passing a fake that conforms to the expected interface:

```javascript
Cookies.set('test-cookie', { expires: FakeExpires() });
```

One more option is to allow inserting a custom Date mechanism so that the developer can decide how to handle it. Also, the test can just verify if the function is being called with the desired argument, not if the `Date` is on the right state:

```javascript
Cookies.setDateMechanism((expiresAttributeValue) => {
  return new Date(new Date() * 1 + expiresAttributeValue * 864e+5);
});

...

Cookies.setDateMechanism((expiresAttributeValue) => {
  assert.strictEqual(expiresAttributeValue, 5);
  return defaultDateMechanism;
});
```

FagnerMartinsBrack: I'm starting to wonder if maintaining this `expires: Number` thing is worth it and we should instead deprecate it to allow each developer to define their own way to handle dates, without being coupled to JavaScript `Date` class.

carhartl: It is worth rethinking the approach, no question, though of all the options I still find the native `Date` the most compelling one, due to its simplicity (no need for extra implementation, no need for a user to learn a new concept). 

Though I also find that having both `expires: Number` _and_ `expires: Date` is a duality we should be removing in a future version, simply because `Date` is powerful enough to easily cover the "days" scenario. In the past this was also causing confusion (e.g. why do we only support days, but not say hours or minutes?). Supporting `Date` and only `Date` would remove this ambiguity. The number == days case seemed sufficient and the simplest concept back in the jQuery days when the plugin had been initially written.. even support for `Date` was only added later on, when it was realized that not everyone would want to write cookies with expiration periods that could be represented in whole days. Allowing for fraction of days later was wrong imo (I find it rather unintuitive to specify `0.5` if I would want 12 hours until expiration), and was also further blurring the line between `Date` and `Number`...

Regarding comprehensibility of the test, I kind of agree, but whatever mechanism we would be building, in the end we'd have to test that the string being passed to `document.cookie` would contain the expiration in the expected format, I don't see any way around this (think integration test); that is, without rewriting more parts of the library with testability in mind (not against it, we could do this in version 3, also tackling the modularization). The way the code is structured still leans to the old jquery days and maybe it's time to overcome this.

All of this imo is beyond the scope of this supposedly tiny PR, where the aim was to fix a broken test.

carhartl: As mentioned, I would deprecate `expires: Number`, but keep `expires: Date`, thus not invent a new concept for how to deal with a date.

carhartl: PS: The test seems no more or less comprehensible than it was before, it's just fixed :)

carhartl:
..at least now trying to avoid the more cryptic code within the test

test/tests.js
FagnerMartinsBrack: Cool LGTM, maybe something to rethink later.

FagnerMartinsBrack:
Yeah, in the tests we don't have any reason to spare bytes so we can be more verbose if that makes the test clearer

carhartl:
=> 3fcefa52affa80e3febf43352569aa2592af3e60

PR #430, opened 2018-04-18
update README

kimixu717:


FagnerMartinsBrack:
Sorry @kimixu717, we missed this PR. Somebody else created [another one](https://github.com/js-cookie/js-cookie/pull/432) with a better description and that was landed.

PR #437, opened 2018-04-28
Update to node 10 in .travis.yml

FagnerMartinsBrack:


DanielRuf:
In general Node.js 8 and 10 should be added.

FagnerMartinsBrack:
@DanielRuf Why, is that because of LTS?

DanielRuf:
See `npm test`. Dependencies are always relevant as packages can break on newer Node.js releases so js-cookie could not be build there or run at all. CI is for testing if there are any breaking changes in any code (thirdparty or the project code).

As we have LTS 8 and stable 10 which get support and bugfixes, only the are relevant. 4 is EOL and 6 also is not the current LTS and is out of its most lifecycle phases.

FagnerMartinsBrack:
I see, but that's for development purposes only. We can just restrict the development process to use only one specific supported node version. I don't see value in having the development process to support multiple versions.

One way to document the supported node version is to add a `engines` property to the `package.json`. What do you think?

DanielRuf:
> but that's for development purposes only

That's what CI is for. To check commits, pushes and PRs.

DanielRuf:
> We can just restrict the development process to use only one specific supported node version

This is not advised / a good idea and will make things harder for contributors.

We for example use Node 10 by default for all projects.

FagnerMartinsBrack:
Fair enough, it's not like we need a lot of effort to support node versions for development other than add a new line to the yaml. @DanielRuf want to create a PR with all the node versions you believe are worth supporting for local development?

DanielRuf:
> I don't see value in having the development process to support multiple versions

These are checks to make sure that everything is fine on all machines with different Node.js versions.

When I added Node.js 10 to many projects, also many of them showed issues on Node.js 10 to specific (outdated) dependencies which really broke it on the machines.

DanielRuf:
> One way to document the supported node version is to add a engines property to the package.json. What do you think?

This is also not good.

Best example: upath 1.0.4 is used by webpack and was set to Node <= 9. It broke setups on Node.js 10. 1.0.5+ fixed this and removed the engine restriction.

Even old and well written code will work on Node.js 10. Virtual barriers and conflicts caused by `engine` would abort `npm i`, also when a package is a dependency of another package like upath.

See also https://github.com/ai/size-limit/pull/61

FagnerMartinsBrack:
Cool, feel free to create a PR upgrading to node 8 and 10.

.travis.yml
```diff
@@ -3,6 +3,7 @@ node_js:
   - '6'
 # Only use grunt-ci for commits pushed to this repo. Fall back to regular test
 # for pull requests (as secure variables won't be exposed there).
+  - '10'
```
FagnerMartinsBrack: This is even, wrong, the code was added below a comment that should be above the `script` property...

DanielRuf: This is not wrong per se ;-)

DanielRuf: But as you already closed the PR I can not push any change to this PR anymore.

PR #444, opened 2018-05-28
chore: cache node_modules

DanielRuf:


.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: node_js
 node_js:
   - '6'
+cache:
+  directories:
+    - node_modules
```
FagnerMartinsBrack: How's the cache invalidated?

DanielRuf: See https://docs.travis-ci.com/user/caching/#How-does-caching-work%3F

DanielRuf: Probably similar to CircleCI and others who create checksums.

FagnerMartinsBrack: `node_modules` is roughly 150mb and the [link above](https://docs.travis-ci.com/user/caching/#How-does-caching-work%3F) says:

> If you store archives larger than a few hundred megabytes in the cache, it is unlikely that you’ll see a big speed improvement.

Not sure what that means, but let's test if we can get a meaningful change in build time. It probably will.

DanielRuf: You can easily test this by triggering a new build with an empty commit.

FagnerMartinsBrack:
Nice... it looks like there's some Cloudflare problem when trying to access NPM registry: https://travis-ci.org/js-cookie/js-cookie/jobs/385142811#L456. It could be a temporary issue.

PR #445, opened 2018-05-28
chore: clone last 5 commits

DanielRuf:


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
   - '6'
+git:
+  depth: 5
```
FagnerMartinsBrack: What's the difference between the default depth of `50` and using 5 instead?

DanielRuf: `5` is often faster

DanielRuf: Cloning 5 vs 50 commits makes a big difference in commit changes / diffs.

FagnerMartinsBrack: How fast are we talking about compared to the whole build? Milliseconds or seconds?

DanielRuf: Just milliseconds so probably not worth it.

PR #446, opened 2018-05-29
chore: add Node.js 8 and 10

DanielRuf:


carhartl:
I started to wonder, why are we running tests in several node versions instead of just say the latest LTS, given we're (exclusively?) targeting client-side/browser environments... is this to ensure build scripts compatilibity?

FagnerMartinsBrack:
We don't support minified files in NPM, so it makes sense to ensure the build process works across node versions.

PR #451, opened 2018-06-16
Switches linting to eslint & removes jshint and jscs

realityking:


FagnerMartinsBrack:
Good stuff. What's the main reason for the change, though?

realityking:
Sorry, should have given more context. jscs is unmaintained - the repo is archived - and this merges two configs into one, making things a bit easier to grasp.

FagnerMartinsBrack:
I see this looks good then assuming that the behavior remains unchanged.

package.json
```diff
@@ -35,12 +35,11 @@
     "grunt": "1.0.2",
     "grunt-compare-size": "0.4.2",
     "grunt-contrib-connect": "1.0.2",
-    "grunt-contrib-jshint": "1.1.0",
     "grunt-contrib-nodeunit": "2.0.0",
     "grunt-contrib-qunit": "2.0.0",
     "grunt-contrib-uglify": "2.3.0",
     "grunt-contrib-watch": "1.1.0",
-    "grunt-jscs": "3.0.1",
+    "grunt-eslint": "^20.2.0",
```
FagnerMartinsBrack: Just a small comment, can we use fixed versions here?

realityking: Sorry, hadn't noticed you're using fixed versions for dependencies. Fixed now.

realityking:
Behaviour should be the same, I used https://github.com/brenolf/polyjuice for the conversion + couple of manual fixes for the cases it didn't cover.

Issue #415, opened 2018-03-16
Why is the api function responsible for read and write?


nanndemoiikara:
Why is the api function responsible for read and write?
api function any intention to change responsibility depending on the number of arguments?

FagnerMartinsBrack:
In the eyes of the Public API, it makes no difference, you still have read and write methods. Internally, it uses the same function for the purpose of reducing the byte size and allow the code to be GZipped around the 900 bytes target, [as stated in the README as a feature](https://github.com/js-cookie/js-cookie/tree/e56c99086300e0f3bc81d4e85e355a6f0284be77#javascript-cookie---).

This is a compromise between clean code vs 100% code coverage + downloadable bytes.

carhartl:
It is also due to heritage of once being _jquery-cookie_. In jQuery back then it was a common pattern to have this kind of overload for read + write operations.

FagnerMartinsBrack:
An interesting side-effect of this is https://github.com/js-cookie/js-cookie/issues/417.

FagnerMartinsBrack:
I don't think there's anything we can act on here. Closing then...

PR #454, opened 2018-07-01
Split api() into separate functions to get and set cookies.

realityking:
After reading #415 I wanted to try myself what impact splitting the internal implementation into separate get and set methods what have. It actually saves 5 bytes when gzipped!

   raw     gz Compared to eslint @ 94845aa2d9c48d6ea782dd5c84b218ef48a491bb
    +8     -5 build/js.cookie-2.2.0.min.js
   -34    -28 src/js.cookie.js

I suspect there's also a tiny performance benefit as the number and type of arguments is more predictable for the JIT compiler.

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -36,6 +36,10 @@
 		return result;
 	}
 
+	function decode (s) {
+		return s.replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent);
+	}
+
 	function init (converter) {
 		function api() {}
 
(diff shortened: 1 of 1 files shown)


src/js.cookie.js
```diff
@@ -136,17 +139,15 @@
 			return key ? jar[key] : jar;
 		}
 
-		api.set = api;
+		api.set = set;
 		api.get = function (key) {
-			return api.call(api, key);
+			return get(key, false);
```
FagnerMartinsBrack: [Boolean Trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)?

FagnerMartinsBrack: Maybe because it's internal function we can use a constant to name the argument. Should have no big deal in byte size because minifiers should be able to strip the constant out.

realityking: Unfortunately Uglify - at least the version used here - didn't inline the variable so I added the comment instead.

FagnerMartinsBrack:
That's cool! 866b to 861b, good work!

screen shot 2018-07-03 at 10 02 18 pm

PR #456, opened 2018-07-03
[RFC] Generate CommonJS and ESM versions of JSCookie along with the UMD version.

realityking:
Caveats up front: this is not quite done (but working) and would obviously require a major release. Together with #370 it could make for a nice major release.

The idea here is to keep the source code as an ESM and then use rollup to generate ESM, CommonJS and UMD modules. The major benefit is that the CJS and ESM versions allow users who use Webpack/Rollup/Parcel to save some bytes and some function invocations on startup, which improves startup time.

The UMD build generated by rollup is 9 bytes large then what's currently in master. I think that's an acceptable tradeoff considering how much everyone who uses a bundler saves.

I've included the ESM version in this PR, I'd back it out before merging (or at least not include it in the `package.json`) as the introp between ESM default exports and CommonJS is still not there.

Things that'd are left do:

- [ ] Update docs
- [ ] Remove ESM version
- [ ] Write changelog (maybe)
- [ ] Integrate with grunt watch (maybe)
- [ ] Make sure Travis rejects PRs that didn't update `lib/`

WDYT?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,5 +30,11 @@
     "no-array-constructor": "error",
     "no-new-object": "error",
     "no-new-wrappers": "error"
-  }
+  },
+  "overrides": [{
+    "files": ["rollup.config.js"],
+    "parserOptions": {
+      "sourceType": "module"
+    }
+  }]
 }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -42,6 +42,7 @@ module.exports = function (grunt) {
 		},
 		eslint: {
 			grunt: 'Gruntfile.js',
+			rollup: 'rollup.config.js',
 			source: 'src/**/*.js',
 			tests: ['test/**/*.js', '!test/polyfill.js']
 		},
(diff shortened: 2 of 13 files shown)


FagnerMartinsBrack:
I challenge the value of having different files for different module types. Can we keep the code written in ES5 and implement ES6 in the sidelines so that whoever wants ES6 can use it? Can we do it in a way we don't need to maintain 2 files or add dependency, even in itself?

Some people might use [rawgit](https://rawgit.com/) and link to the file on `master`. I know that is not encouraged but still, I would argue it's good not to have those things broken unless we really need to.

Ping @carhartl

realityking:
Rollup supports going from a CommonJS module to an UMD, that'd be possible. But the code generated is actually a little bit less compact then if it's generated from an ESM. AFAIK no tool supports going from an UMD to anything else.

Note that all the files in `lib/` don't really need to be maintained, rollup just rebuilds them whenever you make a change to the file in `src/`. Though it's probably a good idea to add a test to Travis that this happened before merging a PR.

TBH if users are using rawgit _with the master branch_, nothing could ever be changed without running the risk of it breaking. That's why both unpkg and rawgit recommend to use it based on tags. Personally I'd just let it break but if it's a concern, the file structure could be changed to switch `lib` and `src` around.

As for motivation, I'm trying to eliminate UMD module from my build. They're not just bigger then necessary when using a bundler, they also affect start up time and prevent static analysis for IDE autocompletion and some bundler features (tree shaking and module concatenation).

Now admittedly js-cookie doesn't get the static analysis benefit due to the way the code is structured but IMHO it's a step in the right direction.

FagnerMartinsBrack:
After looking at this again, believe adding more code at this point is not worth it. When the browser support for ESM is big enough, then we might just change the default export to use ESM. Also, this PR has a lot of changes and it will require a review bandwidth this project doesn't have.

I recommend creating a fork and use it in your own project. We don't intend to make too many changes in the code, so keeping in sync shouldn't be that hard.

In the meantime, we'll watch the progress on native modules. Right now the syntax works in many mainstream browsers and TypeScript. In the future, we'll probably dump the conditionals on top and only support ESM.

Thanks for the PR and sorry it couldn't land.

PR #466, opened 2018-09-12
problem with decode regexp

fnnzzz:
Hi, I found a problem in case, when to me from the back-end comes a cookie query-string like, where a part of the value is encoded in Cyrillic and is presented in lower case.

I could not understand why the `get()` method returns `undefined`.

A little debugged made the situation clear - RegExp used in decodeURIComponent matches only lowerCase parts, therefore inside `decode()` function is throw an error `URI malformed` (because of an incorrect string).

image

So I send you my PR with a simple solution to this problem.

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -37,7 +37,7 @@
 	}
 
 	function decode (s) {
-		return s.replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent);
+		return s.replace(/(%[0-9A-Z]{2})+/gi, decodeURIComponent);
 	}
 
 	function init (converter) {


mac2000:
And where is test to catch regression in future?

mac2000:
Just to clarify problem exists in .net implementation of url encode - https://stackoverflow.com/questions/918019/net-urlencode-lowercase-problem

And RFC https://tools.ietf.org/html/rfc3986 says:


The uppercase hexadecimal digits 'A' through 'F' are equivalent to
   the lowercase digits 'a' through 'f', respectively.

FagnerMartinsBrack:
@mac2000 Thank you for the research!
@fnnzzz Thanks for the very clear report!

The RFC says:

> For consistency, URI producers and
   normalizers should use uppercase hexadecimal digits for all percent-
   encodings

But it also says:

> The uppercase hexadecimal digits 'A' through 'F' are equivalent to
   the lowercase digits 'a' through 'f', respectively.  If two URIs
   differ only in the case of hexadecimal digits used in percent-encoded
   octets, they are equivalent.

I can see a coupe of options:

1. Accept lowercase percent-encoding data to make it possible for backend systems to integrate with the standard js-cookie encoding algorithm. We still need a test for it as @mac2000 pointed out.
2. Do not accept lowercase percent-encoding data. Document somewhere that if you have a custom server, we recommend the user to use a [converter](https://github.com/js-cookie/js-cookie/tree/bf8bb9cb99ec9b1aa6e8ee4594823ebc6507d44c#converters) instead and leave the standard js-cookie encoding/decoding algorithm only for compliant servers or to/from js-cookie interchange.

Be aware that option 1 might have an edge-case regression for people that rely on `undefined` for this kind of behavior.

Which one should we go for?
Ping @carhartl

carhartl:
Fixed: 18ed0fd05ae87f11906ae53545681d36731806e6

PR #467, opened 2018-09-23
Updated the README for deleting with a domain

### LINKED ISSUE #463: Can't delete Google Analytics cookies.
Hello,

It may be due to my unability to use your tool even as it seems pretty straightforward.

I want to remove all cookies left over after blocking GA. My .js file contains following code:

```
window['ga-disable-UA-XXXXXXXX-X'] = true; // This blocks GA service, works properly.
var DelCookies = Cookies.noConflict();
var cookies    = DelCookies.get();

if ( '' !== cookies ) {

	jQuery.each( cookies, function( index ) {

		index = '' + index;

		if ( '_ga' === index || '_gid' === index || '_gat' === index ) {

			DelCookies.remove( index, { path: '/' });
		}

	});

}
```

But cookies are still there when I check Chrome's developer console. I used '/' as domain path. I've found this path in the developer console as well.

Best regards, Dan.

ed-parry:
Resolves #463 by updating the README to include an example on deleting a cookie and setting the domain value. Re-wrote the prepending paragraph to lead into the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,11 @@ Cookies.remove('name'); // fail!
 Cookies.remove('name', { path: '' }); // removed!
 ```
 
-*IMPORTANT! When deleting a cookie, you must pass the exact same path and domain attributes that were used to set the cookie, unless you're relying on the [default attributes](#cookie-attributes).*
+*IMPORTANT! When deleting a cookie and you're not relying on the [default attributes](#cookie-attributes), you must pass the exact same path and domain attributes that were used to set the cookie:*
+
+```javascript
+Cookies.remove('name', { path: '', domain: '.yourdomain.com' });
+```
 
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
@ed-parry Can you use the same e-mail of your github account in the commit? You can do so by using the command:

```
$ git commit --amend --author="Ed Parry<my-email@example.com>"
```

This way, the commit will show your avatar when I land it to master.

Thanks.

ed-parry:
I amended the commit and pushed it: https://github.com/js-cookie/js-cookie/pull/467/commits/a78d20b7e5df98040ce57b98477838110cac1d8b so would that not do it, or should I now be removing the previous version? Sorry, git not my strong point beyond the basics!

FagnerMartinsBrack:
@ed-parry that's fine :)

Issue #470, opened 2018-10-05
Cookies.set silently fails if max content size is exceeded


TomSmartBishop:
In Firefox I saw that the following call silently fails:
Cookies.set("test", aStringWithMoreThan4094bytes, { expires: 365 });

Or maybe I just don't know how to check for that?

Would there be a way to create a second cookie and store more?

FagnerMartinsBrack:
See https://github.com/js-cookie/js-cookie/wiki/Frequently-Asked-Questions#why-are-my-cookies-being-deleted. Why are you storing 4092 bytes in a cookie? Be aware they are sent in the headers of every request without GZip optimization.

TomSmartBishop:
I was just prototyping, now using localStorage instead. But that was not the point, I think it just shouldn't fail silently. Cheers, Thomas

FagnerMartinsBrack:
@TomSmartBishop In the spirit of feedback, what do you suggest to do in this specific case? Would having a `js-cookie.debug.js` file with warning and logs help to make things clear?

See also https://github.com/js-cookie/js-cookie/issues/88#issuecomment-140106983 and https://github.com/js-cookie/js-cookie/issues/86#issuecomment-140058592.

TomSmartBishop:
Maybe yes

mauriciabad:
I'm having the same problem as @TomSmartBishop.
I think that the `README.md` file should say that there's a max cookie size in the encoding section.

FagnerMartinsBrack:
Alright, so it seems the wiki page is not enough https://github.com/js-cookie/js-cookie/wiki/Frequently-Asked-Questions#why-are-my-cookies-being-deleted. Can one if you create a PR to improve the README? Perhaps link to the Wiki?

Thanks.

TomSmartBishop:
Just as a comment: I had a single value suddenly exceeding the size, so nothing was 'deleted' just the new value wasn't stored and the old one kept showing up.

FagnerMartinsBrack:
@TomSmartBishop Can you provide an [SSCCE](http://sscce.org/)? You can attach a nodeJS project in a zip file to Github issues.

PR #471, opened 2018-10-13
Remind that cookies may get deleted

mauriciabad:
A fix for issue https://github.com/js-cookie/js-cookie/issues/470


FagnerMartinsBrack:
Thanks @mauriciabad. Landed.

PR #477, opened 2018-11-13
JS-Cookie:Removed PhantomJS Dependency

ossdev07:
PhantomJS is an unmaintained project, Updated grunt-contrib-qunit version to 3.0.1 in package.json to remove phantomjs dependency and use Headless Chrome Instead.

Signed-off-by: ossdev <ossdev@puresoftware.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "grunt-compare-size": "0.4.2",
     "grunt-contrib-connect": "1.0.2",
     "grunt-contrib-nodeunit": "2.0.0",
-    "grunt-contrib-qunit": "2.0.0",
+    "grunt-contrib-qunit": "3.0.1",
     "grunt-contrib-uglify": "2.3.0",
     "grunt-contrib-watch": "1.1.0",
     "grunt-eslint": "21.0.0",


FagnerMartinsBrack:
Last attempt tp update `grunt-contrib-qunit` to version 3.0.0 broke our cross-browser testing suite. See https://github.com/js-cookie/js-cookie/pull/462. Since then it's pending to look at and upgrade.

We'll close this in favor of the older PR. Thanks for letting us know anyway :)

ossdev07:
Hi @FagnerMartinsBrack 

Thanks for taking the time to look at this PR.
I understand your concern but I would like to add that as you can check with the latest version of grunt-contrib-qunit ie 3.0.1 the package is working fine. It will be very helpful if you may consider this. 
Also I checked the older PR #462 too and with a quick look at the travis-ci tests I think the package was failing due to not using of --no-sandbox flag .
I will be happy to help if you need any.

FagnerMartinsBrack:
> I will be happy to help if you need any.

It would be great to understand why CI needs the "--no-sandbox" flag. Travis/Sauce Labs was working for a long time before it and didn't require the "--no-sandbox" flag. Maybe research at potential changes in "Sauce Labs/Travis CI"  that caused the breakage?

In order to reproduce in your own, you can create a SauceLabs or Travis account in a fork and try to run it in a PR just like this project does it. Right now the cross browser testing is configured to not run in PRs but you might be able to run with your own Sauce Labs key.

That would help immensely. If you have time to look at it, feel free to comment your findings on https://github.com/js-cookie/js-cookie/pull/462 :)

ossdev07:
Hi @FagnerMartinsBrack 

> It would be great to understand why CI needs the "--no-sandbox" flag

-> Reason for using --no-sandbox flag:

For aarch64 chrome binary is not present, we are using chromium binary to launch the Chrome.
Security is one of the most important goals for Chromium, and --no-sandbox flag provides that.
When trying to use chromium as root user like in docker container then only the requirement of --no-sandbox flag is coming
Without the flag, chromium gives error as root user-

```
[28769:28769:0925/163702.505790:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
```

Thus, we need to give the --no-sandbox flag when we launch chrome with chromium binary.

Please do revert if you still find anything not clear.

FagnerMartinsBrack:
I think this is valid but seems to be low priority. We only use those dependencies in the dev environment. I'll keep it open for awareness but I believe this will be fixed anyway one we upgrade grunt-contrib-qunit

carhartl:
Closed via eaa734aa1360b05e6d3b9196ab20aa7d1c0b30c0

PR #522, opened 2019-07-24
More lax on supported expires types

mems:
Fix the error when the expires is a number as string

### DIFF
--- a/src/js.cookie.js
+++ b/src/js.cookie.js
@@ -52,8 +52,12 @@
 				path: '/'
 			}, api.defaults, attributes);
 
-			if (typeof attributes.expires === 'number') {
-				attributes.expires = new Date(new Date() * 1 + attributes.expires * 864e+5);
+			// Check Date validity
+			if (attributes.expires instanceof Date) {
+				attributes.expires = isNaN(Number(attributes.expires)) ? '' : attributes.expires;
+			} else if (attributes.expires) {
+				var expires = parseFloat(attributes.expires);
+				attributes.expires = isNaN(expires) ? '' : new Date(new Date() * 1 + expires * 864e+5);
 			}
 
 			// We're using "expires" because "max-age" is not supported by IE


carhartl:
We are already coercing the expires attribute like so: `attributes.expires * 864e+5` so I guess  changing the if condition from `if (typeof attributes.expires === 'number')` to `if (!(attributes.expires instanceof Date))` would be so sufficient. 

I’m just not sure relaxing the input type is a desirable thing, after all it’s documented and therefore expected (expectable?).

src/js.cookie.js
```diff
@@ -52,8 +52,12 @@
...
 			}, api.defaults, attributes);
 
-			if (typeof attributes.expires === 'number') {
-				attributes.expires = new Date(new Date() * 1 + attributes.expires * 864e+5);
+			// Check Date validity
+			if (attributes.expires instanceof Date) {
+				attributes.expires = isNaN(Number(attributes.expires)) ? '' : attributes.expires;
+			} else if (attributes.expires) {
+				var expires = parseFloat(attributes.expires);
+				attributes.expires = isNaN(expires) ? '' : new Date(new Date() * 1 + expires * 864e+5);
```
FagnerMartinsBrack: Need a failing test to justify the code change, see https://github.com/js-cookie/js-cookie/blob/d2724a87a29cbf7ee539452fb42b3e3a1aff95a8/CONTRIBUTING.md#pull-requests

carhartl: I didn’t even mention lack of tests because I’m not sure we want to relax the api in the first place.

FagnerMartinsBrack: That's right, agree on that

FagnerMartinsBrack:
Closing due to the lack of feedback from the [OP](https://medium.com/@fagnerbrack/what-does-op-means-on-github-8db958b604af). If you want to continue the discussion, please comment here mentioning one of the project members.

Issue #535, opened 2019-09-06
Drop support for Bower 


carhartl:
https://www.npmjs.com/package/bower

> We don't recommend using Bower for new projects. Please consider Yarn and Webpack or Parcel. You can read how to migrate legacy project here: https://bower.io/blog/2017/how-to-migrate-away-from-bower/

carhartl:
Closed via #537

Issue #528, opened 2019-08-24
Can I help modernize the package?


mkay581:
First, let me say that I really love this package! But there are a few shortcomings:

1. it isn't interoperable with third-party bundlers like webpack
1. doesn't have built-in typescript support
1. doesn't support ES6 syntax

Instead of creating yet another cookie lib, I'd really like to add this feature set to this package which could help a lot of the developers who are expecting these features. Is this something you all are open to?

carhartl:
Hi Mark, these are welcomed suggestions. Some of these things I had in mind for a fully modernized v3 release of the plugin, where we’d be dropping support for old browsers. Though before I am willing to begin any of that we need to fix automated testing in at least those browsers we would want to support there, see #524.

I’m still not convinced about TypeScript support though, we have had requests and discussions about in the past. We could start without though.

What would be required to support webpack?

mkay581:
Sure, TypeScript is optional. But it should at least have a type declaration file which would make the package interoperable when importing this package into a project that uses TypeScript files.

What browsers did you plan to drop in v3? 

> What would be required to support webpack?

I believe adding support for bundlers like Rollup and Webpack is just declaring the ESM version of the package in the package.json `module` field. Webpack talks about it [here](https://webpack.js.org/guides/author-libraries/#final-steps) and Rollup talks about it [here](https://github.com/rollup/rollup/wiki/pkg.module).

This is to prepare for using ES modules in node. [Here is an awesome writeup](https://medium.com/@nodejs/announcing-a-new-experimental-modules-1be8d2d6c2ff) for what's to come.

carhartl:
Ah, I've been reading the following the other day: https://philipwalton.com/articles/using-native-javascript-modules-in-production-today/

I find it appealing to migrate our codebase to one or more than one ES module while modernizing. On the other hand I always found it to be an advantage to not require any additional build step for packaging up the code one way or the other; maybe these days that's no longer that relevant, because either browsers support them natively or webpack etc. is a commodity in almost all projects.

The idea I had is that we could be providing separates modules for reading + writing cookies, and another "meta" module combining the two and giving us the well-known `Cookies` based api. Not sure, if it's that's really needed/useful, could be suffering from "wanting too much".

> What browsers did you plan to drop in v3? Oh, actually I see #370 lists that support is being dropped for IE 7, 8, and 9?

Yes, at least. Elsewhere it was suggested to also drop support for IE 10 even if I remember correctly.

[...]

Issue #501, opened 2019-03-27
Please provide a compressed package of es6's module syntax in NPM.


zhangchenglin:
Not everyone will choose a CDN.

FagnerMartinsBrack:
@zhangchenglin or anybody else coming to this issue: Do you have a suggestion in how we can easily provide a minified version with the least amount of code and effort in the current codebase?

To give some context I had the plan to create a release process but eventually dropped that because it was consuming too much of my time.

Please read this for context: https://github.com/js-cookie/js-cookie/issues/57#issuecomment-450299638

Ping the upvoters: @erralb @muuvmuuv @Chmylov @aleqsio @ValterJunior @akkis @lloydjatkinson

muuvmuuv:
@FagnerMartinsBrack for most of my project I use [Rollup](https://rollupjs.org/), see: https://github.com/muuvmuuv/vanilla-click-outside

It is easy to set up and does this job very well.

carhartl:
I’m planning to provide this library as an ES module with a fully modernized v3 release.

carhartl:
With #539 in place, as soon as we release the next version, the release will contain a minified version.

carhartl:
We've released version 3 beta, and the package comes with minified versions, for instance via jsdelivr:

https://cdn.jsdelivr.net/npm/js-cookie@beta/dist/js.cookie.min.js

There's a ES module variant as well:

https://cdn.jsdelivr.net/npm/js-cookie@beta/dist/js.cookie.min.mjs

FagnerMartinsBrack:
@carhartl I added this last line to the release:

> From < 900 bytes gzipped to < 800 bytes gzipped

Issue #423, opened 2018-03-31
Document correct ES6 module usage to avoid "SyntaxError: import not found: default"


FagnerMartinsBrack:
Probably test in a browser that supports ES6 without using any library and document accordingly.

kitak:
@FagnerMartinsBrack 

I tried the contents of https://github.com/js-cookie/js-cookie/issues/426 , but it seems that it will not be solved by `import * as Cookies from './js-cookie-master/src/js.cookie.js';`.
The Cookies variable is Module object.

<img width="564" alt="2018-04-18 16 20 01" src="https://user-images.githubusercontent.com/228410/38917022-676b9eb0-4324-11e8-8640-ee62006c6c1d.png">

If object's name is not `Cookies`, I can refer to` window.Cookies` and it will work.
I think that this project needs another file that is ESModule target. I'll create p-r.

FagnerMartinsBrack:
It seems documenting ES6 is more trouble than it's worth. I reverted the commit. Let's use this issue to test on browsers without extra dependencies and document accordingly, without assuming that the developer is using another third-party tool.

Revert commit: https://github.com/js-cookie/js-cookie/commit/d8b016be0f3c068ee4be8a121313de90502da2e0

rapeflower:
TypeError: p.a.setCookie is not a function(…)

DoubleCouponDay:
![image](https://user-images.githubusercontent.com/7555725/40595222-b51a94ec-6276-11e8-8a5d-01ad471fe653.png)

Clearly not the case. js.cookie.js is a self invoking closure which adds the module "Cookie" to the global namespace. 

>It seems documenting ES6 is more trouble than it's worth.

Fine but dont mislead your users by claiming this is an AMD module / commonjs module.

FagnerMartinsBrack:
@DoubleCouponDay This thread is about ES6 module, which was reverted from the README, not AMD or CommonJS.

We support Asynchronous Module Loaders (like require.js) and CommonJS specification, see the tests for Node.js [here](https://github.com/js-cookie/js-cookie/blob/8b311b9dab5751fab3b09b58d215262e908e92e1/test/node.js) and require.js [here](https://github.com/js-cookie/js-cookie/blob/8b311b9dab5751fab3b09b58d215262e908e92e1/test/amd.js).

The sentence says "JavaScript Cookie can also be loaded as an AMD or CommonJS module". What's misleading there?

DoubleCouponDay:
[...]

PR #537, opened 2019-09-06
Migrate package/source to ES module(s)

### LINKED ISSUE #536: Migrate to ES modules

carhartl:
This is the first part (foundation) of the effort to modernize the package. We start providing the package as an ES module.

While the package's `main` remains a pointer to a CommonJS UMD module we're also providing the ES module variant via the package's `module` property, so that bundlers are aware of it.

Syntax hasn't been modernized yet (out of scope of this PR); it will require adding Babel to the Rollup.js based build tooling for browser compatibility. Similarly, only in a follow-up PR I'm going to provide an automated release process, so we can produce the required distributions in a fail-safe manner.

(Along the way I've updated QUnit "accidentally", this wasn't strictly required for this PR.)

- [x] Turn source into ES module
- [x] Adapt build tooling to produce nomodule/module variants
- [x] Adapt tests to include testing the module
- [x] Adapt documentation
- [x] Adapt package.json (include `main` for commonjs/nomodule variant, and `module`, distributing minified versions via `files` would take care of #501)
- [x] Remove support for bower
- [ ] ~Drop support for `noConflict()`~

Notes:

- Here's a spike being worked on: https://github.com/js-cookie/js-cookie/tree/spike/es-module
- Adapting the code to make full use of ES syntax likely out of scope -- we will need to add babel to the [rollup](https://rollupjs.org/) based build tooling

See #528

Closes #536, #535, #423


carhartl:
> include `main` for commonjs/nomodule variant, and `module`

Both variants should sit in a `dist` directory, for this to ensure correctness we'd need to have an automated release process!

https://www.npmjs.com/package/release-it

carhartl:
Oh, it seems that rollup has an option for noConflict:
```
--noConflict                Generate a noConflict method for UMD globals
```

carhartl:
@mkay581 Do you want to take a look?

README.md
```diff
@@ -117,18 +141,6 @@ Cookies.remove('name', { path: '', domain: '.yourdomain.com' });
...
-If there is any danger of a conflict with the namespace `Cookies`, the `noConflict` method will allow you to define a new namespace and preserve the original one. This is especially useful when running the script on third party sites e.g. as part of a widget or SDK.
-
-```javascript
-// Assign the js-cookie api to a different variable and restore the original "window.Cookies"
-var Cookies2 = Cookies.noConflict();
-Cookies2.set('name', 'value');
-```
-
-*Note: The `.noConflict` method is not necessary when using AMD or CommonJS, thus it is not exposed in those environments.*
-
```
carhartl: Wait, we can bring back this piece of documentation, and mention this applies to the UMD module..

README.md
```diff
@@ -22,35 +23,61 @@ A simple, lightweight JavaScript API for handling cookies
...
 ```
 
-Or include it via [jsDelivr CDN](https://www.jsdelivr.com/package/npm/js-cookie):
+*Not all browsers support ES modules natively yet*. For this reason the npm package/release
+comes with both an ES module as well as an UMD module variant. Include the module along
+with the fallback to account for this:
 
 ```html
-<script src="https://cdn.jsdelivr.net/npm/js-cookie@2/src/js.cookie.min.js"></script>
+<script type="module" src="/path/to/js.cookie.mjs"></script>
```
mkay581: Hmmm enforcing consumers to use a JS file with a `.mjs` file extension at this level seems a little odd. Would all browsers know how to interpret a .mjs file extension?

carhartl: I didn't mean to force consumers, more like provide an example of how to include the module of that package directly, along with the fallback. In general, browsers do not care about the extension, though it's important a server serves `mjs` files with the proper mime type of `text/javascript`.

Generally I'm assuming the more likely use case for ES modules is to be bundled up, so I might add that here...

carhartl: Or do you mean it's too early to use `mjs` files in general? I was looking at https://v8.dev/features/modules and it seemed the way to go..

FagnerMartinsBrack: It seems the value for outputting a `.mjs` extension is to assist on documentation. Should we stick with one file for simplicity? Is there another reason for having both `.js` and `.mjs`?

FagnerMartinsBrack:
> Syntax hasn't been modernized yet (out of scope of this PR)

What do you mean by this part?

carhartl:
@FagnerMartinsBrack 

> What do you mean by this part?

I was referring to that we're not (yet) using modern ES syntax at all, but I wish to start doing so. E.g. arrow functions, `const` etc.

This PR provides the foundation for that (Rollup.js), we'd need an additional babel transpilation step in order to maintain compatibility with some older browsers I assume (not sure whether IE 10 already supports these newer syntax features).

Or we're going for TypeScript, which by now I'm open to.. 😬

README.md
carhartl: For now I'm siding with the node developers and wish to keep the extension:

> Still, we recommend using the .mjs extension for modules, for two reasons:
>
> 1. During development, the .mjs extension makes it crystal clear to you and anyone else looking at your project that the file is a module as opposed to a classic script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than classic scripts, so the difference is hugely important!
> 2. It ensures that your file is parsed as a module by runtimes such as Node.js and d8, and build tools such as Babel. While these environments and tools each have proprietary ways via configuration to interpret files with other extensions as modules, the .mjs extension is the cross-compatible way to ensure that files are treated as modules.

I'm assuming it will be rather rare that the source will be consumed directly, rather than being bundled up with a tool. In the latter case the extension matters less, stressing the aforementioned arguments further.

Let's see how it goes. We can always change that back later on, but let's get some feedback first and start releasing a v3 beta once we're ready...

src/js.cookie.mjs
```diff
@@ -0,0 +1,143 @@
+/*!
```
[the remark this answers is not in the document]
carhartl: See my comment in the other discussion.. for now I'd like to keep it as it is, and be very explicit about this being an ES module.

FagnerMartinsBrack:
I believe TypeScript is an unnecessary build complexity for almost no gain in js-cookie. It's not like we have tons of modules calling each other where an interface becomes strictly necessary to ensure program correctness. Our test suite is pretty thorough, it's not like TypeScript is gonna save us from bugs.

Instead of relying on babel to transpile the code, what if we make sure that we run the tests on every supported browser to make sure we're not using unsupported features (Browserstack)? That way we avoid adding more build steps, we won't need them. That also allows us to focus on reducing the size of the lib by working directly on the code we control instead of magically generating code we don't control.

The only build steps or tools I see value here without an unnecessary level of complexity are the test runs, the minification step, and the byte size check.

README.md
FagnerMartinsBrack: Sounds good, if it's a convention it doesn't hurt simply following it.

carhartl:
> Instead of relying on babel to transpile the code, what if we make sure that we run the tests on every supported browser to make sure we're not using unsupported features (Browserstack)? That way we avoid adding more build steps, we won't need them. That also allows us to focus on reducing the size of the lib by working directly on the code we control instead of magically generating code we don't control.

I agree with all of that. At the same time, though, using the very same talking points, you could argue that this whole PR is not needed. I believe it's worth it starting to write JS code in a more modern fashion.

carhartl:
Also, using newer syntax idioms will likely produce smaller sizes. It really depends on which browsers we need to target though. I haven't thought that through yet.

FagnerMartinsBrack:
Can't we use the modern syntax straight away now that IE support is gone?

carhartl:
>Can't we use the modern syntax straight away now that IE support is gone?

Not as long as we’re supporting IE 10 + 11:
https://caniuse.com/#search=%3D%3E

FagnerMartinsBrack:
I see. From what I looked at the code of seems we still have both options of using the js file or the ES6 one so ppl can still choose only modern code.

Also might need to notify jsdelivr  about which bundle is the suportted one, it seems they're minifying themselves when a new js-cookie comes out since we don't have minified version in the SCM.

carhartl:
..if I find some more time I’m going to experiment with this, in order to see how much bigger a transpiled source would become. Maybe we can save somewhere else, using for instance `forEach()` and stay below the 900 bytes mark.

carhartl:
Started a small fun experiment with the syntax for the `extend` function...

At the moment it looks like this:
```javascript
function extend () {
  var i = 0
  var result = {}
  for (; i < arguments.length; i++) {
    var attributes = arguments[i]
    for (var key in attributes) {
      result[key] = attributes[key]
    }
  }
  return result
}
```

Let's use some of the newer Array methods, without the need for transpiling, it's all still supported by IE 10+. This doesn't look much smaller though, and also not necessarily more readable, so maybe there's not much to gain:

```javascript
function extend () {
  return [].slice.call(arguments).reduce(function (a, c) {
    Object.keys(c).forEach(function (k) {
      a[k] = c[k]
    })
    return a
  })
}
```

Let's start using arrow functions! This is starting to look better:
```javascript
function extend () {
  return [].slice
    .call(arguments)
    .reduce((a, c) => (Object.keys(c).forEach(k => (a[k] = c[k])), a))
}
```
Still we have to deal with `arguments`. Since we're now in transpile land, why not use `Array.from`? This is starting to look better:

```javascript
function extend () {
  return Array.from(arguments)
    .reduce((a, c) => (Object.keys(c).forEach(k => (a[k] = c[k])), a))
}
```

// Still too much code though, let other's do all the heavy-lifting:
```javascript
function extend () {
  return Object.assign({}, ...arguments)
}
```

// So yeah, at this point the `extend` function is a bit pointless (note that Babel would transpile this adding a polyfill for `Object.assign()`):
```javascript
    attributes = Object.assign(
      {
        path: '/'
      },
      api.defaults,
      attributes
    )
```

FagnerMartinsBrack:
yeah, I think we built the extend() function at the time because Object.assign wasn't supported by the browsers using the script

PR #539, opened 2019-09-12
Automate releasing

carhartl:
Another part of the ongoing effort of modernizing the codebase. Since we now need to provide different variants/distributions of the library (module/nonmodule including minified versions), we aim for fully automating publishing to npm, as to avoid errors during the process.

We start using [release-it](https://www.npmjs.com/package/release-it) for fully automating both publishing to npm as well as creating releases on GitHub. Those GitHub releases do contain the minified sources, thus we're able to address #501.

- [x] Set up release-it
- [x] Adapt readme

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,0 +1,5 @@
+{
+  "github": {
+    "release": true
+  }
+}
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,5 +1,19 @@
 {
+  "git": {
+    "commitMessage": "Craft v${version} release",
+    "requireCleanWorkingDir": false,
+    "tagAnnotation": "Release v${version}",
+    "tagName": "v${version}"
+  },
   "github": {
-    "release": true
+    "assets": ["dist/*.mjs", "dist/*.js"],
+    "draft": true,
+    "release": true,
+    "releaseName": "v${version}"
+  },
+  "hooks": {
+    "after:bump": "npm run dist",
+    "after:release": "echo Successfully released ${name} v${version} to ${repo.repository}.",
+    "before:init": "npm test"
   }
 }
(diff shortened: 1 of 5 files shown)


FagnerMartinsBrack:
My concern with a third party release tool is that history tells (as of Java Maven, some Grunt packages, etc.) that there seems to be no tool that can generalize all release process steps for all projects. At some point, the projects end up having either too many configurations or missing features that force project authors to automate their own process using a higher level language (bash, some node scripts, etc.).

In the case of `release-it`, it may or may not create that problem. `js-cookie` is pretty small so it should be easy to change in the future if we need to. In regards to NPM and NodeJS, they're pretty well integrated, so I guess there's a slim chance that the community follows enough conventions that allow it to fit into `js-cookie` so that we won't need to change the dependency at any time in the future

I guess we'll only know in hindsight 😆

carhartl:
Should we introduce a `CHANGELOG.md` file in the project, instead of crafting release notes for a GitHub release? Either way, this would be the only manual step left in the release process...

(Probably doesn't matter much, we won't be releasing all the time, with many changes..?!)

carhartl:
> At some point, the projects end up having either too many configurations or missing features that force project authors to automate their own process using a higher level language (bash, some node scripts, etc.).

So far `release-it` provided everything I expected nicely and it's easy to use and configure, thus good enough (for me). I try to be a bit pragmatic and really really want to avoid a "not invented here" type of thing going on.

carhartl:
PS: And to me it's an improvement over a manual process in any case. Sorry for not picking up the work that had once been started on this.

FagnerMartinsBrack:
> Sorry for not picking up the work that had once been started on this.

I'm glad you didn't. I'm not even sure that effort would give a meaningful ROI anyway

FagnerMartinsBrack:
Don't forget to move the `latest` tag to the latest release. That way it doesn't break README links relying on `/latest`

PR #549, opened 2019-10-09
Update README.md

Parikshit-Hooda:
English Grammar-wise enhancements.
Neither-nor are used together. The word 'nor' is not used separately to the best of my knowledge.
Checked it with Grammarly which indeed showed an error/warning.


Parikshit-Hooda:
@carhartl  This check fails because of some problem with node.js version. However, I have only made refactors, the PR should merge without a hitch. Please see if this can be let through. If there is something wrong on my part, please let me know as I don't think I am causing this check fail.

carhartl:
Thanks! ❤️

PR #564, opened 2019-12-03
travis: add node 12

nikolas:


carhartl:
Adding 12 is fine, but I’d like to keep all older LTS versions.

carhartl:
Thanks! I will merge it on the 31st of December, when Node 8 reached end of life, haha!

nikolas:
Oh woops, i'll add back in node 8 too.

carhartl:
Thanks for reminding us!

PR #566, opened 2019-12-04
Use converter internally

nreynis:
I made some changes so that the library use it's own `withConverter` API internally. This has several avantages:

* using your own API is always a good idea
* minimize branching, clearer logic
* encode / decode methods are now clearly identified
* the internal converter can be exposed to user
* server-side custom converters can reuse the code (examples in SERVER_SIDE.md are greatly simplified)

### DIFF
--- a/src/js.cookie.mjs
+++ b/src/js.cookie.mjs
@@ -9,11 +9,27 @@ function extend () {
   return result
 }
 
-function decode (s) {
-  return s.replace(/(%[\dA-F]{2})+/gi, decodeURIComponent)
+var rfc6265Converter = {
+  read: function (value, key) {
+    return value.replace(/(%[\dA-F]{2})+/gi, decodeURIComponent)
+  },
+  write: function (value, key) {
+    return encodeURIComponent(value).replace(
+      /%(23|24|26|2B|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g,
+      decodeURIComponent
+    )
+  }
 }
 
 function init (converter) {
+  // normalize converter
+  if (!converter.read) {
+    converter = {
+      read: converter,
+      write: converter.write || rfc6265Converter.write
+    }
+  }
+
   function set (key, value, attributes) {
     if (typeof document === 'undefined') {
       return
@@ -28,12 +44,7 @@ function init (converter) {
       attributes.expires = attributes.expires.toUTCString()
     }
 
-    value = converter.write
-      ? converter.write(value, key)
-      : encodeURIComponent(value).replace(
-        /%(23|24|26|2B|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g,
-        decodeURIComponent
-      )
+    value = converter.write(value, key)
 
     key = encodeURIComponent(key)
       .replace(/%(23|24|26|2B|5E|60|7C)/g, decodeURIComponent)
@@ -82,9 +93,8 @@ function init (converter) {
       }
 
       try {
-        var name = decode(parts[0])
-        jar[name] =
-          (converter.read || converter)(cookie, name) || decode(cookie)
+        var name = rfc6265Converter.read(parts[0])
+        jar[name] = converter.read(cookie, name) || rfc6265Converter.read(cookie, name)
 
         if (key === name) {
           break
@@ -110,10 +120,11 @@ function init (converter) {
         })
       )
     },
-    withConverter: init
+    withConverter: init,
+    rfc6265Converter: rfc6265Converter
   }
 
   return api
 }
 
-export default init(function () {})
+export default init(rfc6265Converter)
--- a/test/node.js
+++ b/test/node.js
@@ -24,5 +24,12 @@ exports.node = {
     Cookies.remove('anything')
     Cookies.remove('anything', { path: '' })
     test.done()
+  },
+  shouldExposeRfc6265Converter: function (test) {
+    test.expect(2)
+    var Cookies = require('../dist/js.cookie.min.js')
+    test.ok(!!Cookies.rfc6265Converter.read)
+    test.ok(!!Cookies.rfc6265Converter.write)
+    test.done()
   }
 }
--- a/SERVER_SIDE.md
+++ b/SERVER_SIDE.md
@@ -29,27 +29,13 @@ setrawcookie($name, rawurlencode($value));
 
 ```javascript
 var PHPCookies = Cookies.withConverter({
-  write: function (value) {
-    // Encode all characters according to the "encodeURIComponent" spec
-    return (
-      encodeURIComponent(value)
-        // Revert the characters that are unnecessarily encoded but are
-        // allowed in a cookie value, except for the plus sign (%2B)
-        .replace(
-          /%(23|24|26|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g,
-          decodeURIComponent
-        )
-    )
-  },
+  write: Cookies.rfc6265Converter.write,
   read: function (value) {
-    return (
-      value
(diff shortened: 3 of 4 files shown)


nreynis:
Oh sh... Node download failed and travis went all in and decided to rebuild v8 from the source... This will take hours.

carhartl:
Very nice, thanks! It solves the one mayor shortcoming that we had with the converter implementation — that you couldn’t compose a custom one with the default.

carhartl:
Would it make sense to put the default rfc6265 converter to `defaults` instead of its own property?

carhartl:
Ah no... not so much I guess! When you get to change the defaults an instance of the api has already been created, and then that‘s not very intuitive.

nreynis:
I initially tried to make a named export (kind of thing that work in webpack) but rollup complained that I shouldn't mix `named` and `default` export in the same module.

So I played it safe and exposed it at a property of the API. Another way would be to put it in it's own module, but that would require changes to the build process and I'm not fluent enough with `rollup` to make thoses.

carhartl:
There‘s only a small nitpick, I‘ve thought about the name: `RFC6265Converter`... it looks a bit like a constant, but isn‘t. Would we use `rfc6265Converter`? Or should we actually denote it as a constant (not sure if that‘s idiomatic JavaScript): `RFC6265_DEFAULT_CONVERTER`?

nreynis:
Well js do have constants now and they are usually in screaming snake case (https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Instructions/const). Keep in mind that the const keyword is for variables declaration only, it won't help freezing the api object property. Only talking about the naming convention.

I prefer `rfc6265Converter`, but that's your call.

carhartl:
> I prefer `rfc6265Converter`, but that's your call.

+1

test/node.js
```diff
@@ -24,5 +24,12 @@ exports.node = {
     Cookies.remove('anything')
     Cookies.remove('anything', { path: '' })
     test.done()
+  },
+  should_expose_rfc_6265_converter: function (test) {
```
carhartl: I noticed that we were using snakecase here (might have been added by a pull request and we didn't pay attention) and now changed that: 9834ceb0de76d29629d6c69c4611587326e4043d

Could you adapt this? Hat tip for noticing and keeping it consistent in here 🙇

carhartl:
I've removed the String casting as suggested, needs to be taken care of here as well as to not accidentally bring it back: dc37d14d7a4c99d102fc0acca95a4e3ff440676b

nreynis:
I rebased the branch on master and squashed all the commits in one:
- Changed `RFC6265Converter` to `rfc6265Converter`
- Made sure to not reintroduce string casting removed in dc37d14
- Changed naming convention in tests according to 9834ceb

carhartl:
I just had the thought that we maybe want to expose the rfc6265/standard converter as just `converter` or `defaultConverter` in the api object. The rest, that it’s converting according to rfc6265 seems more like an implementation detail. Not sure though, and can always be done later, after merging.

"converter" + "withConverter" reads better together and seems more consistent to me.

src/js.cookie.mjs
```diff
@@ -82,9 +93,8 @@ function init (converter) {
       }
 
       try {
-        var name = decode(parts[0])
-        jar[name] =
-          (converter.read || converter)(cookie, name) || decode(cookie)
+        var name = rfc6265Converter.read(parts[0])
+        jar[name] = converter.read(cookie, name) || rfc6265Converter.read(cookie, name)
```
carhartl: More of note to myself: do we still need this particular fallback behavior? In the regular case `converter.read` will be  `rfc6265Converter.read`.. only if a custom read converter would return a falsy value we'd fallback to standard rfc6265 read conversion. If I remember correctly this had been implemented as a mechanism to write a converter that only acts on particular keys and otherwise forwards the value to the standard converter. But now that we're providing access to the standard converter, such a mechanism can easily implemented within the custom converter altogether...:

```javascript
var customReadConverter = (value, key) => {
  if (value === 'special") {
    return value..toUpperCase()
  }
  return Cookies.converter.read(value)
}
```

src/js.cookie.mjs
```diff
@@ -9,11 +9,27 @@ function extend () {
...
 }
 
 function init (converter) {
+  // normalize converter
+  if (!converter.read) {
+    converter = {
+      read: converter,
+      write: converter.write || rfc6265Converter.write
+    }
+  }
```
carhartl: Another note: for v3 we may want to change the `withConverter()` api slightly. I don't see a reason why `withConverter(func)` should turn the given func into the read converter. It could just as well have been the write converter. I'd prefer to make passing the type of converter explicit, seems simpler, not only for us to implement, but in the end also for a user to remember.

This would then allow to simplify this to (I believe):

```javascript
function init (converter) {
  converter = extend(rfc6265Converter, converter);
  ...
```
```

```javascript
converter = ({read: converter, write
```

carhartl:
@nreynis Looks like ready to merge? The failing build on Travis is due to Browserstack not working for PRs, though I thought this line in .travis.yml would have prevented running this stage in the first place: `if: env(BROWSERSTACK_USERNAME) IS present`

src/js.cookie.mjs
nreynis: I agree, this would make the API clearer and would simplify the internal logic.

src/js.cookie.mjs
nreynis: I'm also all in for this, in fact this is the issue I had on my head when [I asked to hold release of 3.0](https://github.com/js-cookie/js-cookie/issues/565)

This is needed to pass thoses tests specifically:
https://github.com/js-cookie/js-cookie/blob/master/test/tests.js#L484
https://github.com/js-cookie/js-cookie/blob/master/test/tests.js#L490

This is particulary problematic from my point of view (typescript user) because it weaken the assumption you can make on return type.
If you pass a read converter of type `<T>` you would assume that the following
`
Cookie<T>{
  get(name: string): T | undefined;
  get(): {[key: string]: T};
}
`
But because of this fallback we have to assume that every get could return an union type `T | string`:
`
Cookie<T>{
  get(name: string): T | string | undefined;
  get(): {[key: string]: T | string};
}
`
It would be really great if this could be avoided.

nreynis:
Yep it's all good for me. I can confirm that browserstack tests pass on my fork.

carhartl:
> Do you want me to rename anything before merging or will you take care of that later?

Can do later... I am still not sure about that thought. What do you think?

Thanks a ton, great stuff!

src/js.cookie.mjs
carhartl: => #570

nreynis:
Your welcome.

For the naming I kinda prefer explicits names, if it was just me I would split `Cookie` and the converter to make something like this
`
src/js.cookie.mjs
src/converter/rfc6265.mjs
src/converter/php.mjs
src/converter/tomcat.mjs
src/converter/jboss.mjs
src/converter/express.mjs
`
And make a simple monolithic build with js.cookie.mjs + rfc6265.mjs for people who don't use bundlers.

src/js.cookie.mjs
carhartl: =>#571

carhartl:
We gained some bytes, let's see how we can make up for that... it seems these two simplifications will at least help.

carhartl:
> And make a simple monolithic build with js.cookie.mjs + rfc6265.mjs for people who don't use bundlers.

After having started with ES modules I had a similar idea, making even read and write separate modules (some people might only need to write?). Might happen in the future, at the moment I don't have time for that and I'm also not yet fully convinced if there was enough value.

src/js.cookie.mjs
FagnerMartinsBrack: 👍 to remove `withConverter(fn)`. We did that at the time because we didn't have a use case to write. Then we have so it becomes pointless.

PR #567, opened 2019-12-05
Drop string casting

nreynis:


nreynis:
Sorry, didn't mean to open this here

PR #593, opened 2020-01-31
Add readme warning about `expires` parameter

jakub-g:
I learnt the described issue hard way (in production), so even if it might be weird to add a note like this to readme, if it prevents another developer from breaking production, it might be worth it :)


carhartl:
Thanks for your effort writing this and creating the PR.

I do think though this is too specific information for an already very long readme, and — unless there was some sort of standard — there should be no expectation whatsoever that APIs in two different packages work the same. I could imagine the wiki is a better place for this kind of note..

Issue #595, opened 2020-02-07
Cookie name with brackets


seahindeniz:
**Is your feature request related to a problem? Please describe.**
Because of escaping characters on the key, I can't able to add brackets in key without having escaped characters.
https://github.com/js-cookie/js-cookie/blob/b1d83a073525e00f87d8b4effa4804b8cb741f12/src/api.mjs#L21-L23

**Describe the solution you'd like**
I'm suggesting to have a additional way to override the key replacers same as with the value.
https://github.com/js-cookie/js-cookie/blob/b1d83a073525e00f87d8b4effa4804b8cb741f12/README.md#L308-L312 

to have something like this definition
```js
Cookies.withConverter({
  write: function (value, name) {
    return String(value).toUpperCase()
  }
})
```
and in additional, this definition to have an overloaded return type to keep both definition exist.
```js
Cookies.withConverter({
  write: function(value, key) {
    return {
      key: String(key),
      value: String(value).toUpperCase()
    }
  }
})
```

**Describe alternatives you've considered**
Passing additional optional parameter to disable escapers.
```js
Cookies.set('name', 'value', {
  unescapedName: true,
  unescapedValue: true,
  //...
});
```

**Additional context**
I know this solution can expect this this answer https://github.com/js-cookie/js-cookie/issues/590#issuecomment-575628332 . I'm only suggesting to have an optional way to avoid complying to standards. 

Such as this
https://github.com/js-cookie/js-cookie/blob/b1d83a073525e00f87d8b4effa4804b8cb741f12/src/rfc6265.mjs#L5-L10

carhartl:
Could you explain *why* you need brackets in the cookie‘s key?

seahindeniz:
I'll give you a quick example of my real problem..

I'm currently developing a browser extension for a web site.

![image](https://user-images.githubusercontent.com/5789670/74010534-c6f3a600-4996-11ea-8642-49ba8a420b69.png)
In this screenshot, these cookies are settled by web site and as you can assume, I'm trying to override the value inside of it with this way
```js
cookie.set("Zadanepl_cookie[infobar]", infoBarBase64);
```
First red arrow is the result

nreynis:
I think this feature request is kinda related to this one:  https://github.com/js-cookie/js-cookie/issues/560

carhartl:
@seahindeniz Do you have control over the `Zadanepl_cookie[infobar]` cookie?

seahindeniz:
[...]

PR #608, opened 2020-03-02
Revisit encoding/decoding implementation

### LINKED ISSUE #590: Cookie name with slash
**Describe the bug**
Cookie name with slash `/` character is escaped to `%2F`
Eg. `user/123=foo` becomes `user%2F123=foo`

I believe the forward slash character is allowed in the cookie key.

It's unclear why you're passing `key` here as the converter expects a single parameter:
`value = converter.write(value, key)`

And on the next line:
https://github.com/js-cookie/js-cookie/blob/b1d83a073525e00f87d8b4effa4804b8cb741f12/src/api.mjs#L21

Perhaps it would be better to have:
`key = converter.write(key)`

**Expected behavior**
Cookie name preserves the original forward slash character.

carhartl:
Our RFC 6265 based cookie name + value encoding was based on the
requirements for server implementers, and as such too strict.

We need to look at RFC 6265 section 5.2, requirements for user agents.

But even the algorithm described in section 5.2 isn't that relevant
for us, in that we had to follow the steps when writing a cookie with
this library. This is left to the browser vendors who have to implement
the setter functionality behind `document.cookie`. All we have to do is
avoid surprises when said algorithm is being followed while
`document.cookie = ...` is being executed with the cookie string we're
constructing.

It means we have to encode ";" and "=" in the cookie name, and ";" in
the cookie value.

Follow-up to discussion in #595.

Closes #590, #595.

~Open question: should we be liberal in what we accept/read, that is apply `decodeURIComponent()` to the cookie value, in case we get a server-side written cookie that adheres to RFC 6265? Since we will use percent encoding for ";" in the value we could support that without a lot of effort. An advantage might be that this way we'd be reading v2 cookies without a problem, making the transition from v2 -> v3 not require a cookie migration. Or would we leave that to the user who would have to implement a converter for a complete percent decoding?~

The upside is that we're again well-below 700kb for the gzipped size of the library.

PS: I would probably also want to rename the "rfc6265" converter to something else now as part of this... basically they're semicolon converters now.


carhartl:
>Should we be liberal in what we accept/read, that is apply decodeURIComponent() to the cookie value when reading, in case we get a server-side written cookie that adheres to RFC 6265?

Wrong! Same mistake again. This is making the assumption that servers adhering to RFC 6265 would use percent-encoding to encode the forbidden characters mentioned in the spec, but the spec doesn't say anything about how those characters should be encoded. It mentions Base64 in fact:

>servers that wish to store arbitrary data in a cookie-value SHOULD encode that data, for example, using Base64

carhartl:
So.. we are going to apply the minimal encoding/decoding, for everything else we are providing converters.

carhartl:
I have no idea why the pr check fails in Travis, while the one for push passes...

FagnerMartinsBrack:
Need to verify if [SERVER_SIDE.md](https://github.com/js-cookie/js-cookie/blob/d727ba6ceffb101e300a38b4eb06bcca4ac571d1/SERVER_SIDE.md) fixes still hold true

README.md
```diff
@@ -166,12 +165,9 @@ _Note: The `.noConflict` method is not necessary when using AMD or CommonJS, thu
...
-The only character in cookie-name or cookie-value that is allowed and still encoded is the percent `%` character, it is escaped in order to interpret percent input as literal.  
+Special characters that are not permitted in the cookie name (";" and "=") or cookie value (";") are encoded with their UTF-8 Hex equivalent using [percent-encoding](http://en.wikipedia.org/wiki/Percent-encoding).
 Please note that the default encoding/decoding strategy is meant to be interoperable [only between cookies that are read/written by js-cookie](https://github.com/js-cookie/js-cookie/pull/200#discussion_r63270778). To override the default encoding/decoding strategy you need to use a [converter](#converters).
 
-_Note: According to [RFC 6265](https://tools.ietf.org/html/rfc6265#section-6.1), your cookies may get deleted if they are too big or there are too many cookies in the same domain, [more details here](https://github.com/js-cookie/js-cookie/wiki/Frequently-Asked-Questions#why-are-my-cookies-being-deleted)._
```
FagnerMartinsBrack: Why removing this?

Gruntfile.js
```diff
@@ -1,33 +1,8 @@
...
-}
-
 const config = {
   qunit: {
     all: {
       options: {
-        urls: [
-          'http://127.0.0.1:9998/',
-          'http://127.0.0.1:9998/module.html',
-          'http://127.0.0.1:9998/encoding.html?integration_baseurl=http://127.0.0.1:9998/'
```
FagnerMartinsBrack: Need also to remove [this part](https://github.com/js-cookie/js-cookie/blob/d727ba6ceffb101e300a38b4eb06bcca4ac571d1/CONTRIBUTING.md#integration-with-server-side) from the README which documents the `?integration_baseurl`.

FagnerMartinsBrack:
> The upside is that we're again well-below 700kb for the gzipped size of the library.

Can change the README to <700kb gzipped then! :D

README.md
carhartl: Still true, but imo in the wrong place, it doesn't really have anything to do with encoding.

carhartl: (And after all, there's an entry in the FAQ...)

carhartl:
>Need to verify if SERVER_SIDE.md fixes still hold true

I looked at these and I think so.. or they've been wrong even before these changes here: for instance what confused me was that the docs suggest to escape parens for Tomcat:

```javascript
var TomcatCookies = Cookies.withConverter({
  write: function (value) {
    return (
      Cookies.converter
        .write(value)
        // Encode the parens that are interpreted incorrectly by Tomcat
        .replace(/[()]/g, escape)
    )
  },
  read: Cookies.converter.read
})
```

but the default converter was percent-encoding `(` and `)` anyway.

README.md
carhartl: Ok to resolve?

FagnerMartinsBrack: Sure

PR #625, opened 2020-04-30
Fix typo in docs

NicoPennec:
According the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) and the official [RFC](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05), the SameSite value must be written with  first character capitalized (as done in your [unit test](https://github.com/js-cookie/js-cookie/blob/master/test/tests.js#L15)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,14 +253,14 @@ Cookies.remove('name')
 
 ### sameSite
 
-A [`String`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String), with possible values `lax` or `strict`, prevents the browser from sending cookie along with cross-site requests.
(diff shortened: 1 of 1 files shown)


carhartl:
Thanks for the PR!

For the cookie http header it is true, but in `document.cookie` case doesn‘t matter. The MDN docs are using all lowercase:

https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie

"LAX", "lax", "Lax" will all work..

carhartl:
The specification itself requires case-insensitive matching:

> If "cookie-av"'s "attribute-value" is not a case-insensitive
       match for "Strict" or "Lax", ignore the "cookie-av".

https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00#section-4.1

PR #630, opened 2020-05-21
Updated documentation with sameSite example

xeno:
Found this example buried in a closed issue and figured it was worth adding to the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,12 @@ Create an expiring cookie, valid to the path of the current page:
 Cookies.set('name', 'value', { expires: 7, path: '' })
 ```
 
+Create a cookie with SameSite attributes:
+
+```javascript
+Cookies.set('name', 'value', { secure: true, sameSite: 'lax'})
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,6 +106,12 @@ Create an expiring cookie, valid to the path of the current page:
 Cookies.set('name', 'value', { expires: 7, path: '' })
 ```
 
+Create a cookie with SameSite attributes:
```
carhartl, line 109: I think it should say “... SameSite attribute” (no plural, there’s only one such attribute).

carhartl: Ah, it refers to the additional `secure` flag I guess.

carhartl:
There’s an entire section regarding the SameSite attribute, along with an example, that should be sufficient. The examples are not meant to provide an exhaustive list of usages.

See: https://github.com/js-cookie/js-cookie/blob/master/README.md#samesite

Thank‘s for the PR, but I‘m going to close it for this reason.

PR #643, opened 2020-07-10
To prevent overwriting cookie in same key when calling Cookies.get()

Arman19941113:
Example in `xx.yy.com`:

```js
document.cookie = 'demo=1;domain=xx.yy.com'
document.cookie = 'demo=2;domain=yy.com'
console.log(document.cookie ) // 'demo=1; demo=2'
```

Excepted:

```js
Cookies.get('demo') === Cookies.get().demo
```

In fact:

```js
Cookies.get('demo') === '1'
Cookies.get().demo === '2'
```

This behavior may be confused.

### DIFF
--- a/src/api.mjs
+++ b/src/api.mjs
@@ -51,7 +51,10 @@ function init (converter, defaultAttributes) {
       var parts = cookies[i].split('=')
       var value = parts.slice(1).join('=')
       var foundKey = defaultConverter.read(parts[0]).replace(/%3D/g, '=')
-      jar[foundKey] = converter.read(value, foundKey)
+      // To prevent overwriting cookie in same key
+      if (!jar[foundKey]) {
+        jar[foundKey] = converter.read(value, foundKey)
+      }
 
       if (key === foundKey) {
         break


carhartl:
Did you write a test for this? I'm not aware that `document.cookie` when accessed as a reader will be providing a string containing a duplicate key. What I know is that a cookie with different attributes (say `path`) can shadow another one with the same name, so depending on the current context, you get one or the other.

carhartl:
So at the moment we will be returning the last cookie that was found for a given key. This PR changes that to become the first. I don’t see why that would be a better behavior, both ways it is sort of random which one we get.

Arman19941113:
Sometimes the priority of cookie is useful.
If a domain has cookie like this `'demo=1; demo=2'`.
I think`Cookies.get('demo')` should equal to `Cookies.get().demo`.
Now this is not equal.
I think it may be unreasonalble.

Arman19941113:
'a.b.com' will share cookies in 'b.com'. So 'a.b.com.cookie' may contain a
duplicate key.

Klaus Hartl <notifications@github.com> 于2020年7月10日周五 下午5:18写道：

carhartl:
>Sometimes the priority of cookie is useful.

What defines the priority though? Is it the order in which cookies have been written (if that determines in which order cookies are returned by `document.cookie`). We can't tell when reading the cookie which domain/path it is from, so we can't deduce a priority from that. And is the order, in which cookies are returned by `document.cookie` the same in all the browser we support, or could one browser possibly return `'demo=2; demo=`'` - I will want to more thoroughly analyze this.

>I think Cookies.get('demo') should equal to Cookies.get().demo

I agree, but they shouldn't be equal in a possible broken or inconsistent way.

I am closing this PR (we don't accept PRs without tests anyway), and create a ticket to track this problem instead.

Arman19941113:
Without setting `path` attribute, The priority is not sure.
In safari recent modified cookie has high priority. But chrome is just the opposite.
This pr is mean to let `Cookies.get('demo') === Cookies.get().demo` when we have duplicate cookie names.

PR #644, opened 2020-07-10
Optional to get last cookie when matching many cookies in same key

Arman19941113:
Example in `xx.yy.com`:

```js
document.cookie = 'language=en;domain=yy.com'
document.cookie = 'language=zh-cn;domain=xx.yy.com'
console.log(document.cookie ) // 'demo=en; demo=zh-cn'

console.log(Cookies.get('language')) // 'en'

// If I do this:
document.cookie = 'language=de;domain=xx.yy.com'
console.log(document.cookie ) // 'demo=en; demo=de'

// I will always get 'en'
console.log(Cookies.get('language')) // 'en'
```

Sometimes we want the last cookie:

```js
Cookies.get('demo', 'last') === 'de'
```

### DIFF
--- a/src/api.mjs
+++ b/src/api.mjs
@@ -38,7 +38,7 @@ function init (converter, defaultAttributes) {
     return (document.cookie = key + '=' + value + stringifiedAttributes)
   }
 
-  function get (key) {
+  function get (key, position) {
     if (typeof document === 'undefined' || (arguments.length && !key)) {
       return
     }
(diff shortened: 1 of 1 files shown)


carhartl:
We might observe more than two cookies with the same name in a particular page/path, so this won't really solve the actual problem, thus I'm closing. (We don't accept PRs without tests anyway.)

Issue #646, opened 2020-07-13
Using js-cookie on third party sites / noConflict


unuseless:
The bug happens in the following circumstance:

A site uses requirejs for loading js (called parent site).  In this parent site third party javascript snippets are used.  The third party scripts (i.e. js based polls, popups, counters, ...) have no access to the requirejs loading environment of the parent site.

js-cookie is used in such a third party script.  Because the third party script does not know in which environment js-cookie is loaded, they should use the `noConflict` method.  That works well if the parent site does not use requirejs.  However if the parent site uses requirejs, the `noConflict` method throws an error because it is not exported.

This issue is very similar to #234 .

**Expected behavior**
js-cookie allows `noConflict` under all circumstances, even if the host site uses requirejs.

carhartl:
@unuseless You have an idea how to solve this?

unuseless:
@carhartl What I've done to circumvent the problem is: editing the code of js-cookie like this:
```
/*!
 * JavaScript Cookie v2.2.1
 * https://github.com/js-cookie/js-cookie
 *
 * Copyright 2006, 2015 Klaus Hartl & Fagner Brack
 * Released under the MIT license
 */
;(function (factory) {

    // 2020.07.13 edited to remove check for module loader, as in
    // third party script must be able to load through
    // global object


    // var registeredInModuleLoader;
    //     if (typeof define === 'function' && define.amd) {
    //     	define(factory);
    //     	registeredInModuleLoader = true;
    //     }
    //     if (typeof exports === 'object') {
    //     	module.exports = factory();
    //     	registeredInModuleLoader = true;
    //     }
    //     if (!registeredInModuleLoader) {
    var OldCookies = window.Cookies;
    var api = window.Cookies = factory();
    api.noConflict = function () {
	window.Cookies = OldCookies;
	return api;
    };
    //	}
}(function () {
....
```
i.e. commenting out the check for the module loader which disables the `noConflict`. That even makes the code smaller :-)

carhartl:
🤔

We‘re currently having rollup generate this piece of code for us during build.

carhartl:
Also, it removes support for loading js-cookie as a UMD/AMD Module, so that‘s clearly not an option for us.

unuseless:
An alternative is:
```
 //     if (!registeredInModuleLoader) {
    var OldCookies = window.Cookies;
    var api = window.Cookies = factory();
    api.noConflict = function () {
	window.Cookies = OldCookies;
[...]

Issue #580, opened 2019-12-12
Cookies.noConflict() not working


wwestonh:
Starting late last night (12/11/2019) I noticed that the method noConflict() was no longer recognized using the CDN call:

<script src="https://cdn.jsdelivr.net/npm/js-cookie@beta/dist/js.cookie.min.js"></script>
<script type="module" src="https://cdn.jsdelivr.net/npm/js-cookie@beta/dist/js.cookie.min.mjs"></script>

Error:

"Uncaught TypeError: Cookies.noConflict is not a function"

carhartl:
If you're including the module version this doesn't make sense. If you need `noConflict()`, don't use the module:

```
<script src="https://cdn.jsdelivr.net/npm/js-cookie@beta/dist/js.cookie.min.js"></script>
```

carhartl:
Probably caused by https://github.com/js-cookie/js-cookie/commit/47b538d69f8fd36e821015ab4ce8dec6e0a122e6

carhartl:
Bumped the beta and pushed a new version to npm.

wwestonh:
Will the jsdelivr CDN version at https://cdn.jsdelivr.net/npm/js-cookie@beta/dist/js.cookie.min.js be updated?

carhartl:
I don’t know. I think so, it’ll pick up whatever the tag “beta” is pointing to on npm.

PR #666, opened 2020-10-04
Generate CommonJS and ESM versions of JSCookie along with the UMD version.

realityking:
This is quite complex makes sure every environments gets the best possible module:

- unpkg and JSDelivr get a minified UMD module
- Newer Node.js versions get an ESM modules, older a CommonJS module
- The same is true for bundlers, if the bundler is configured to use ESMs it will receive it. Otherwise a CommonJS module will be used.

The use of `browser` in the old package.json is problematic as many bundles, including webpack, will by default prefer it over `module`. In that case the user ends up with an already minified UMD. The UMD is much worse, size wise, than the CommonJS version. Minified dependencies are also bad for debuging.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -46,7 +46,8 @@ const config = {
       'dist/js.cookie.mjs',
       'dist/js.cookie.min.mjs',
       'dist/js.cookie.js',
-      'dist/js.cookie.min.js'
+      'dist/js.cookie.umd.js',
+      'dist/js.cookie.umd.min.js'
     ],
     options: {
       compress: {
--- a/package.json
+++ b/package.json
@@ -2,8 +2,14 @@
   "name": "js-cookie",
   "version": "3.0.0-rc.1",
   "description": "A simple, lightweight JavaScript API for handling cookies",
-  "browser": "dist/js.cookie.js",
+  "main": "dist/js.cookie.js",
   "module": "dist/js.cookie.mjs",
+  "unpkg": "dist/js.cookie.umd.min.js",
+  "jsdelivr": "dist/js.cookie.umd.min.js",
+  "exports": {
+    "import": "dist/js.cookie.mjs",
+    "require": "dist/js.cookie.js"
+  },
   "directories": {
     "test": "test"
   },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -21,11 +21,16 @@ export default [
       },
       // config for <script nomodule>
       {
-        file: pkg.browser,
+        file: pkg.unpkg.replace('.min.js', '.js'),
         format: 'umd',
         name: 'Cookies',
         noConflict: true,
         banner: ';'
+      },
+      // config for older Node.js versions
+      {
+        file: pkg.main,
+        format: 'cjs'
       }
     ],
     plugins: [licenseBanner]
@@ -40,7 +45,7 @@ export default [
       },
       // config for <script nomodule>
       {
-        file: pkg.browser.replace('.js', '.min.js'),
+        file: pkg.unpkg,
         format: 'umd',
         name: 'Cookies',
         noConflict: true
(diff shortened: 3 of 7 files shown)


carhartl:
In the meantime I started contemplating - for v3 - to drop support for umd modules altogether:

https://github.com/js-cookie/js-cookie/issues/646#issuecomment-696086251

We shouldn’t even need CommonJS modules, given the lib isn’t supposed to and will just not work in a node environment. To make this explicit, we have chosen to use the `browser` field for the npm package.

Didn’t think through all the implications of dropping support for cjs yet, certainly it would require ES modules aware bundlers, which I‘m fine for v3. For everything else there would always be v2.

realityking:
I think I'd agree on dropping UMD. AMD/RequireJS seems thoroughly dead and probably not worth serving anymore.

I do very much disagree on CommonJS. Native ESM module are still in its infancy in many ways. I think it's premature to abandon CommonJS. For example, we had to disable loading ESMs in Webpack because some packages released ESM modules incompatible with the CommonJS modules in the same package which broke either the web build or the Node.js process. Rollup makes it pretty painless to support both and while the number of special case `package.json` fields is rather annoying it does create a very seamless experience.

carhartl:
So how could a solution look like when dropping support for amd, i.e. umd? We'd have to start providing an _iife_ variant to be used for users not using a bundler (realistic?) and in order to solve #646. 

Ideally I'm thinking we'd have three variants (iife -> dist/js.cookie.js, esm -> dist/js.cookie.mjs, cmd -> dist/js.cookie.cjs):

```
"browser": "dist/js.cookie.js"
"exports": {
  "import": "dist/js.cookie.mjs",
  "require": "dist/js.cookie.cjs"
}
```

(or `main` instead of `browser` if latter that's an issue with bundlers).

carhartl:
Hmmm, no longer building a umd variant we'd loose `noConflict()` which apparently people are using: #580 (I guess when referencing the file directly from cdns). Not sure whether this would be an acceptable breaking change for v3, but I'm thinking it's not.

carhartl:
I'm so confused: https://github.com/js-cookie/js-cookie/issues/646#issuecomment-689666615

CyberFlameGO (approved):
LGTM!

carhartl (approved):
Going to merge..

carhartl:
Thanks @realityking 😍

carhartl:
@realityking Let‘s say I need to add another iife variant for dist, is there a predefined/recommended setup for package.json?

(Needed to close #646)

realityking:
Thanks for merging! I don't fully understand your question. What variant are you trying to add?

carhartl:
We can try this again without breaking jsdelivr 😬

realityking:
I didn’t realize people were so frequently deep linking to a specific build file and don’t pin to specific versions. How annoying (and what an anti pattern when packages can declare which files are to be used for jsdeliver).

Two options:
1. Maintain B/C bu using a slightly different naming scheme so that the UMD files keep their existing names.
2. Avoid using the old UMD names altogether and force break this entirely to make people learn.

Just kidding about 2. I’ll make a PR for approach 1 this weekend.

carhartl:
Unfortunately we (the maintainers) created this problem ourselves, by advertising the deep link in the readme: dd1699f8f834fb9a083a1fffaba80fc3bf9c2a32

Despite prominent note (never prominent enough) I believe many developers are quickly copy&pasting the link from the readme in master, ending up with a release candidate. 💥 

I have removed the example altogether now, you get more convenient copy&paste links on jsdelivr anyway.

carhartl:
Thanks for creating another PR. I've been thinking about the support for older Node.js versions (the `cjs` format that was added to rollup) - which node versions are these? I'm not too keen on having this, to me the `browser` field really makes sense for this library. E.g. it would be acceptable for me to restrict it to newer versions for a v3 release. On the other hand, I experienced myself that Webpack requires some extra config to disrespect the `browser` field, and that's also not ideal.. a bit torn here.


```js
// make web pack skip package.json “browser” field
module.exports = {
  resolve: {
    mainFields: ['module', 'main']
  }
}
```

PR #695, opened 2021-07-14
fix `exports` field path

jacklesliewise:
The `exports` field contains paths that are incorrect according to [Node.js documentation](https://nodejs.org/api/packages.html#packages_exports).

This throws a `ERR_INVALID_PACKAGE_TARGET` error: https://nodejs.org/api/errors.html#errors_err_invalid_package_target

This PR fixes the issue and bumps the version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,14 @@
 {
   "name": "js-cookie",
-  "version": "3.0.0-rc.2",
+  "version": "3.0.0-rc.3",
   "description": "A simple, lightweight JavaScript API for handling cookies",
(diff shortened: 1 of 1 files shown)


hendrikmolder (approved):

carhartl:
Thank you @jacklesliewise 😍

carhartl:
(Did own commit, since we do the bumping via release-it.)

jacklesliewise:
Awesome, thank you @carhartl - nice work!

PR #727, opened 2021-08-31
export package.json

odeniso:
```
rollup-plugin-svelte: The following packages did not export their `package.json` file so we could not check the `svelte` field. If you had difficulties importing svelte components from a package, then please contact the author and ask them to export the package.json file.
```

[here](https://github.com/sveltejs/rollup-plugin-svelte/issues/181) is some more context from within the svelte project

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,11 @@
   "unpkg": "dist/js.cookie.min.js",
   "jsdelivr": "dist/js.cookie.min.js",
   "exports": {
-    "import": "./dist/js.cookie.mjs",
-    "require": "./dist/js.cookie.js"
+    ".": {
+      "import": "./dist/js.cookie.mjs",
+      "require": "./dist/js.cookie.js"
+    },
+    "./package.json": "./package.json"
   },
   "directories": {
     "test": "test"


carhartl:
It‘s about https://github.com/sveltejs/rollup-plugin-svelte#pkgsvelte

carhartl:
I‘d rather not litter the package.json with non-standard fields, and it seems it isn‘t strictly necessary according to aforementioned documentation.

lukeed:
No, it's about the "package.json" not being accessible to the outside world. Doesn't matter who or what is trying to read the package.json file, anything outside of this module's own files can't look at the file cuz it's not defined as an export.

This is very standard & has been part of the [ESM resolution algorithm](https://nodejs.org/api/esm.html#esm_resolver_algorithm_specification) and [official documentation + example](https://nodejs.org/api/packages.html#packages_package_entry_points) since very early days.

odeniso:
Thanks for the quick look @carhartl!

As @lukeed mentions above, this is not introducing the non-standard field (I assume you are referring the the `"svelte"` property). It is merely exposing the package's `package.json` file, which is often assumed to be accessible:

> Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. require('your-package/package.json'). This will likely be a breaking change. 
https://nodejs.org/api/packages.html#packages_package_entry_points

carhartl:
Thanks for the PR, and apologies for taking too quick a look..

lukeed:
Thanks, appreciate it!

carhartl:
@odeniso @lukeed I've created a new release so the fix becomes available: https://github.com/js-cookie/js-cookie/releases/tag/v3.0.1

PR #771, opened 2022-04-07
docs: use https protocol for RFC 6265

kpman:
The `http` protocol will cause `not found` in the first render. Using `https` will be better.


carhartl:
Thank you ❤️

PR #800, opened 2022-10-19
remove all user visible cookies 

tazimmadre:


carhartl:
Thank you for the PR and your contribution. But! This has been discussed before and the answer was and still is that we don‘t want to add this functionality to the library rather than providing the building blocks for something like this to be implemented. We don’t want to provide an abstraction for a loop.

Also see https://github.com/js-cookie/js-cookie/wiki/Frequently-Asked-Questions#how-to-remove-all-cookies

PR #807, opened 2023-02-12
Expose all lib variants from dist

BendingBender:
- add all built variants in ./dist to package.json's export field
- add a .npmrc to disable package-lock.json generation

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
--- a/package.json
+++ b/package.json
@@ -9,7 +9,16 @@
   "exports": {
     ".": {
       "import": "./dist/js.cookie.mjs",
-      "require": "./dist/js.cookie.js"
+      "browser": "./dist/js.cookie.js",
+      "default": "./dist/js.cookie.js"
+    },
+    "./dist/js.cookie": {
+      "import": "./dist/js.cookie.mjs",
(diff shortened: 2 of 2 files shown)


carhartl:
Btw, every few years I'm contemplating removing the min variant from the dist altogether - I have no numbers, but my hunch is that manually downloading the lib and referencing the file from your web server is a common scenario. And for those who need to do this, referencing the lib from a publicly available CDN would even be advantageous.

BendingBender:
The way it's currently set up, node won't allow the usage of these bundles, it will error out. Only bundlers that don't follow node policy will allow it, I'm not sure whether webpack/skypack/... will allow it. So it's kinda half broken already. You can of course decide to remove the bundles, it's up to you.

carhartl:
>The way it's currently set up, node won't allow the usage of these bundles, it will error out. Only bundlers that don't follow node policy will allow it, I'm not sure whether webpack/skypack/... will allow it.

Understood! My ~~thinking~~assumption though is that when using a bundler like webpack you won't have a need for the minified dist in the first place, e.g. I would (finally) leave minification to the users of the lib.

carhartl:
(Providing a minified version as part of the dist really to me is a remnant of a feature couple of years ago where bundlers were less common and users asked for such a version often.)

PR #881, opened 2025-08-22
Note about `HttpOnly`

olfek:


daviddanielng (commented):
It is well know that httponly cookie can not be accessed with JavaScript.

It is a simple Google search.

daviddanielng:
@olfek 
There is a reason this library do not have `httpOnly` attribute when setting cookies with `Cookie.set()` .

How can one forget what Httponly cookies are? That's like forgetting what `string` and `int` are.

The only use of httponly cookie is to make it inaccessible from the client. i.e JavaScript.

daviddanielng:
@olfek 
>An attempt to block this is just shameful.

I am not a member of this library neither do my opinion count, the change is a plus.

I am just saying it is redundant, almost everyone setting httponly cookies knows it can only be accessed by server.

daviddanielng:
> More incorrect assumptions that affect some portion of users responsible for ~13 million downloads weekly. And a very careless attitude for such a popular library.

Dev error, happens all the time .

Even if the message was plastered everywhere, people will still run into this problem.

Example (ME) : [https://github.com/remix-run/react-router/issues/14202](issue at react router)

PR #940, opened 2026-04-23
debugging Browserstack

carhartl:


FagnerMartinsBrack:
Aaaarhgh I was very confident this would fix it: https://github.com/js-cookie/js-cookie/commit/9f37cecfc756bf194b57b894596fe150d701e076

Looks like every iOS real-mobile device is not working, probably an issue on BrowserStack (bs-local seem to be used only for iOS devices)

carhartl:
@FagnerMartinsBrack Given that the `browserstack-runner` package is basically unmaintained since many years, and that there are [open bugs](https://github.com/browserstack/browserstack-runner/issues/201) that might or might not be what we're experiencing here (hard to tell) I started migrating the testing approach over to using the Browserstack SDK (with the help of Codex). It seems to look good, [I ran a POC for iPhone](https://github.com/js-cookie/js-cookie/actions/runs/24925685600/job/72995161513)... I will be doing some more testing and cleanup when merging this.

carhartl:
Stopping this approach..

carhartl:
I attempted to use the `bs-local.com` domain [earlier](847c8923f6d3d330e9515a806f53e418ea09b4e2), but this would only work with [your change](https://github.com/js-cookie/js-cookie/commit/9f37cecfc756bf194b57b894596fe150d701e076) I guess, @FagnerMartinsBrack 

I don't recall at all why we have `"browserstack.local": "false"` in the configuration.

carhartl:
...

FagnerMartinsBrack:
> I attempted to use the `bs-local.com` domain [earlier](847c8923f6d3d330e9515a806f53e418ea09b4e2), but this would only work with [your change](https://github.com/js-cookie/js-cookie/commit/9f37cecfc756bf194b57b894596fe150d701e076) I guess, @FagnerMartinsBrack
> 
> I don't recall at all why we have `"browserstack.local": "false"` in the configuration.

If you look at the history it was added as part of a commit to add the ios tests last time. It was there since day 1. If that started breaking afterwards then it's definitely a browserstack bug

carhartl:
> I attempted to use the `bs-local.com` domain [earlier](847c8923f6d3d330e9515a806f53e418ea09b4e2), but this would only work with [your change](https://github.com/js-cookie/js-cookie/commit/9f37cecfc756bf194b57b894596fe150d701e076) I guess

No..

PR #958, opened 2026-05-20
Fix test message typo and replace deprecated `escape()` call.

Mohammad-Faiz-Cloud-Engineer:
Two small, safe fixes I noticed while reading through the codebase:

### 1. Test message typo - `test/encoding.js`
Line 1163 had `should_handle` with an underscore instead of a space. Everything else in the test suite uses natural spacing in assertion messages, so this was just a copy-paste slip. Makes it consistent with the other ~100 messages.

### 2. Removed deprecated `escape()` - `src/api.mjs` The `set` function uses `escape()` to encode `(` and `)` in cookie names. `escape()` has been deprecated since forever and some tooling flags it. Replaced it with explicit `.replace(/\(/g, '%28').replace(/\)/g, '%29')` - same output, no behavior change, one less deprecation warning.

### DIFF
--- a/src/api.mjs
+++ b/src/api.mjs
@@ -18,7 +18,8 @@ function init(converter, defaultAttributes) {
 
     name = encodeURIComponent(name)
       .replace(/%(2[346B]|5E|60|7C)/g, decodeURIComponent)
-      .replace(/[()]/g, escape)
+      .replace(/\(/g, '%28')
+      .replace(/\)/g, '%29')
 
     var stringifiedAttributes = ''
     for (var attributeName in attributes) {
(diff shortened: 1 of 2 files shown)


Mohammad-Faiz-Cloud-Engineer:
I Have Fixed the Issue, Please Check and Merge it into main.

carhartl:
If you want to have this merged, please address the problematic shape of commits.

The initial commit combines two unrelated changes, while the second should have been a fixup and comes with a commit message far from providing useful context for future maintainers.

Say we‘ll have to revert one or the other change. We‘ll have no choice and revert both changes, plus are forced to revert two commits in order. Likewise, we‘re making blame harder to use like this.

carhartl:
More on the topic of atomic commits: https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/

(Atomic PRs would be even more preferred btw…)

carhartl:
PS: the contributing guidelines have that:

> Create a new topic branch for every separate change you make.

But it could be phrased more clearly, going to do that.

Mohammad-Faiz-Cloud-Engineer:
Getting errors while making the changes, so I’m closing this PR for now. I’ll create a new one with the fixes.

PR #965, opened 2026-05-24
Fix: cookie names colliding with Object.prototype properties being silently dropped

Mohammad-Faiz-Cloud-Engineer:
Found a bug in get() - when a cookie is named after an Object.prototype property (like toString, constructor, hasOwnProperty, etc.), the in operator was matching the inherited property before the cookie could be stored in the jar. So the cookie would just disappear, no error, nothing.

Switched to Object.hasOwn() which only checks own properties - now those cookies actually get returned like they should. Also fixed a random typo in a test assertion string while I was at it.

### DIFF
--- a/src/api.mjs
+++ b/src/api.mjs
@@ -61,7 +61,8 @@ function init(converter, defaultAttributes) {
 
       try {
         var found = decodeURIComponent(parts[0])
-        if (!(found in jar)) jar[found] = converter.read(value, found)
+        if (!Object.hasOwn(jar, found))
+          jar[found] = converter.read(value, found)
         if (name === found) {
           break
         }
--- a/src/api.mjs
+++ b/src/api.mjs
@@ -18,7 +18,8 @@ function init(converter, defaultAttributes) {
 
     name = encodeURIComponent(name)
       .replace(/%(2[346B]|5E|60|7C)/g, decodeURIComponent)
-      .replace(/[()]/g, escape)
+      .replace(/\(/g, '%28')
+      .replace(/\)/g, '%29')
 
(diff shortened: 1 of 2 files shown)


carhartl:
Like I mentioned before, please do not send PRs with two separate changes merged into a single commit.

carhartl:
As stated in the contributing guidelines:

> Create a new topic branch for every separate change you make.

Thus, expected here would be one PR for the bugfix, and another one for the typo.

carhartl:
Furthermore, we cannot use `hasOwn()` just so, since we‘re still officially supporting IE 10 + 11 with version 3.

https://caniuse.com/?search=hasOwn

I think these properties clashing is an acceptable edge case we‘re not supporting.

carhartl:
We could just as well change line 57 to

```js
var jar = {__proto__: null}
```

carhartl:
If you could create an issue for this, it‘d be much appreciated. We can then figure out whether we want to fix it, and what solution would yield the smallest size addition while maintaining the promised browser compatibility.. thanks!

Mohammad-Faiz-Cloud-Engineer:
Done Please Check.

carhartl:
The last commit doesn‘t make sense (it reverts a change introduced in the first, the change not belonging to this PR in the first place). This whole PR should consist only of the second commit. It‘s also adding the change related to `escape`, which doesn’t belong here either.

Furthermore, it‘s lacking a test for the behavior you‘re attempting to fix. I‘m closing this PR, we keep the issue.

carhartl:
I will be taking care of the rest.

PR #966, opened 2026-05-24
Fix typo in test assertion string

Mohammad-Faiz-Cloud-Engineer:
Fixes a typo in `test/encoding.js` where `should_handle` (underscore) was used instead of `should handle` (space) in the assertion message for the 4-byte UTF-8 character test.


carhartl:
e7d9a4dff0d5ebc51a200819e3ab3a2dbe87979a

PR #973, opened 2026-05-28
IE testing

carhartl:
- **Reapply "Activate Browserstack testing for PRs"**
- **Run test in IE**


carhartl:
Still the same error, not related to restored ES5 compatibility.

